PR #242, opened 2024-07-09
[Language Support] Turkish layout added 

remreren:
Adds Turkish characters in addition to qwerty layout.

I have added the keys but unable to determine which characters to pop up when long pressed to new characters.

New layout:
image

New layout with number line and hints enabled:
image


abb128:
Merged in 270ff11feae245d0b7ed55f9291e07d0251ba33a, thanks!

PR #398, opened 2024-07-30
Layouts: east_slavic layout update

XeonDead:
Add additional long-press symbols, linked from QWERTY layout as well as additional symbols sprinkled from GBoard layout.
I'm aware of backslash not showing up as hint, but unfortunately I'm not really much into Java, so that I can't help with.

### DIFF
--- a/java/res/xml/rowkeys_east_slavic2.xml
+++ b/java/res/xml/rowkeys_east_slavic2.xml
@@ -23,37 +23,50 @@
 >
     <!-- U+0444: "ф" CYRILLIC SMALL LETTER EF -->
     <Key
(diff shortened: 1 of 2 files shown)


abb128:
Thank you for the PR. The layout system has been updated and symbols are now added automatically instead of each layout needing to define them manually, so this should no longer be necessary. The XML files still remain but they're unused and will be deleted once all of them have been ported to the new system. You can see the new east_slavic definition here https://github.com/futo-org/android-keyboard/blob/master/java/assets/layouts/Slavic/east_slavic.yaml

XeonDead:
Ah, okay, I see, thanks for following up.

PR #473, opened 2024-08-15
Update Bulgarian.java

Bogo25:
Renamed the layout name and removed 'ѝ' from the menu which is displayed when 'и' is long-pressed. I did not find out how to modify uppercase in regards to uppercase 'ь'.


abb128:
It appears you've edited a test for the layout and not the actual layout itself. 

FYI the layout system is undergoing a pretty big change at the moment to make things less confusing, so it might be worth waiting until that's complete before making any layout changes

PR #535, opened 2024-09-08
Improving Typing Support for Devanagari

Midnight1938:
Hindi is a language that has always had atrocious support on android. With some years of switching around, have found SwiftKey's implementation the best out of all (most simply copy what google has).

I am trying to port the same to Futo keyboard, because of its offline features, and have been able to add the layout to the Layouts Picker as a [YAML File](https://github.com/Midnight1938/android-keyboard/blob/master/java/assets/layouts/Hindi/hindi_devanagari.yaml), but do not know the process required to implement it properly.

(ignore version changes as this pull is for review)


Midnight1938:
> Please sign our contributor license agreement at https://cla.futo.org

I did?

tom-futo:
> > Please sign our contributor license agreement at https://cla.futo.org
> 
> I did?

A bug in the CLA bot. It's webhook is subscribed to any changes to the PR and comments. 

I see you have the CLA-signed label so it's good to go!

Midnight1938:
> Thanks for the PR. The new layout system was still WIP at the time you made it but the integration is finished now. To add the layout you just need the yaml file and add `- devanagari` under the target languages in mapping.yaml
> 
> Is there a reason you put `!text/morekeys_symbols_1` on all the numbers? What's the `# doesnt fix it` comment?

Hi, the doesn't fix was for the addition of `!text/morekeys_symbols_1` nit fixing the keyboard crashing.

Should i open a different PR with the latest modifications?

abb128:
You can edit the current PR. I would suggest removing the `!text/morekeys_symbols_1`. Also it's not necessary to define a bottom row, one is added automatically, explicitly defining it is mainly if you need to customize the default.

If you want to disable the default symbols, you can also add `attributes: {moreKeyMode: OnlyExplicit}` above `rows:`. I don't type in these languages myself so I'm not sure if the default long-press symbols are desirable, but if they are then you can just leave it as-is.

You shouldn't experience any crashing on the latest master, if you still do then please post a stack trace if possible

abb128:
Please open a PR on https://github.com/futo-org/futo-keyboard-layouts with the updated version if you'd still like to see it added. The new nightly has a built-in layout editor as well to help test and develop the layout.

Midnight1938:
sorry about that, forgot about the pr due to life

Issue #386, opened 2024-07-28
Make the ñ key hold type a "/"


BoscoDomingo:
### Language

Spanish

### Description

```ñ``` should print "/" when long-pressed, such as in the Google Keyboard:

![Screenshot_2024-07-28-17-20-45-018_com miui notes-edit](https://github.com/user-attachments/assets/e53d367a-f3ff-4651-9dba-cd1270f3fc0c)

### Complexity

0 complexity. Just adding a new symbol to an existing letter

abb128:
This is fixed in nightly

PR #555, opened 2024-09-16
Add "/"  to the additional keys from ñ character in Spanish layout

BoscoDomingo:
Solves #386 .

Code may be completely wrong, I just went for what seemed intuitive to get the ball rolling, but I didn't even build it, let alone test it. If not enough, let me know!


abb128:
This is fixed now in a different way

BoscoDomingo:
> This is fixed now in a different way

@abb128 Fantastic, thank you kindly!

PR #593, opened 2024-10-03
Add Icelandic layout for Icelandic language

ornstrange:
Add Icelandic specific layout for Icelandic language.
Uses same layout as IOS and SwiftKey:

swiftkey

Without number row:
nonumber

With number row:
withnumber


abb128:
Thanks!

PR #922, opened 2024-12-07
Update lv.json

arbitrary-dev:
Move letters absent in latvian language to miscellaneous category.


arbitrary-dev:
I can't sign your CLA. Because this code snipped is stolen and violates several licenses, patents, NDAs and overall is a hate speech on itsown 😄

abb128:
Sorry for the delay, I see that these letters are only historically used, so I've moved them to the misc category

PR #1095, opened 2025-02-04
Fix: remove non-exsisting symbol form UK

SkeLLLa:
Removes non-existing symbol from Ukrainian language.

See https://en.wikipedia.org/wiki/Ukrainian_alphabet for reference.

### DIFF
--- a/tests/src/org/futo/inputmethod/keyboard/layout/tests/TestsUkrainian.java
+++ b/tests/src/org/futo/inputmethod/keyboard/layout/tests/TestsUkrainian.java
@@ -17,7 +17,7 @@
 package org.futo.inputmethod.keyboard.layout.tests;
 
 import android.test.suitebuilder.annotation.SmallTest;
-import java.util.Locale;
+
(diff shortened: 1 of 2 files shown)


abb128:
Thanks for the PR. I've looked into how other keyboards handle it, and some seem to keep the ъ letter as an extra option. I don't really see much reason to remove it, it might help writing Russian words, but I see that it can get in the way of writing the exclamation mark. The misc morekeys exist for this purpose as they commonly contain keys from other languages but have lower priority than symbols by default. I've gone ahead moved the letter into misc

SkeLLLa:
@abb128 I can't agree.

It doesn't help writing russian words, since there are no other russian symbols in the Ukrainian layout. Also even in russian `ъ` sign is very uncommon. However it hides `!` sign (on long press) that is more widespread and used in Ukrainian language.

So UX wise it's not good to put `ъ` as a first option on long press, while `!` is the second and requires more moves to select it.

PR #1097, opened 2025-02-04
more keys for Nepal Bhasa / Newa

sapradhan:
More keys for [Nepal Bhasa](https://en.wikipedia.org/wiki/Newar_language) in [Newa/Prachalit](https://en.wikipedia.org/wiki/Pracalit_script) script.


sapradhan:
CLA signed but not label not updated, please let me know if something needs to be done. 

image

abb128:
Thanks!

PR #1119, opened 2025-02-14
Improve device selection logic for audio recognition

xunxky:
Improve device selection logic for audio recognition in case of non-standard mic types.


xunxky:
Not all people live in Texas and are able to comply with such a broad statement.  It would be great if FUto would acknowledge that fact. (a "paying" customer ... it feels like self-checkout LMAO)

abb128:
@xunxky I am sorry you felt the CLA is too complicated. The CLA does not require you to comply with any significant terms or to live in Texas, it just assigns joint copyright ownership to use your code contributions in our software. For such a minor change it's probably not necessary, but our bot does not have a law degree to tell which contributions would require a CLA and which ones might not. If you open or link to an issue that describes the problem that this fixes I can implement the fix for you

PR #1223, opened 2025-04-06
Sanitize whisper output

s-h-a-d-o-w:
I recently started to heavily use the voice input and enjoying it quite a bit except for the fact that I need to make a lot of minor corrections because of whisper being unaware of text context. Unfortunately, as you probably know, it's not really feasible to add this context to the prompt, since whisper prompts are mostly about style and hints, not context that it should use verbatim.

And so I figured I'd suggest a solution if it can be easily integrated into the current code. I'm not experienced with Android development, so please forgive me if there are some things that aren't idiomatic.

## Example

Using the text "This is the first test." as the basis for each case and the following actions ...

1. Replace "first" with "second".
2. Delete "test" and dictate "thing".
3. Delete "This is" and dictate "We're at".

... the following happens:

Current `master`:

1. This is the Second. test.
2. This is the first Thing..
3. We're at... the first test.

This proposal:

1. This is the second test.
2. This is the first thing.
3. We're at the first test.

## Caveats:

- We can't know whether a result starts with a name. (But I do check for acronyms.)
- We can't tell whether the end of the whisper output actually should be the end of a sentence. So we are only be able to remove excess punctuation, not add any.

Both of these could obviously be resolved by using a general purpose LLM but I suspect that that would add too much delay to voice input processing overall (also quite a bit more complex to implement), so I decided to go ahead with this proposal.
If nothing else, at least maybe this could be a temporary improvement if it is found that using a general purpose model instead is actually feasible.


s-h-a-d-o-w:
Through some weird behavior that I noticed while using this in practice (a space getting added after "(", even though my tests confirm that that shouldn't be done), I realized that I should have put this logic into `java\src\org\futo\inputmethod\latin\uix\UixManager.kt`.

Will migrate the code there in the coming days.
(Even if this shouldn't get merged, I will just use a build of my fork - I really need this. 😄 But if that's the case, I would of course appreciate feedback as to what problems you see with my solution.)

s-h-a-d-o-w:
This is ready to review now, sorry about the premature submission.

abb128:
Thanks!

Issue #1207, opened 2025-04-01
Clipboard History Grid View


Unbaised99:
### Use Cases

I want to request Clipboard History view as Grid 

![Image](https://github.com/user-attachments/assets/f2f376a2-5c32-4029-a5b5-67f7cc5bada0)

As list view is not good Looking 

### References
Florisboard 
Attached SS for Reference
_No response_

PR #1225, opened 2025-04-06
Reduce minimum column size on clipboard history view

venkatesh2090:
#1207 Says a user wants grid view to the clipboard history view. But seems like grid view is already being used. Just reduced the min size for an entry so more items are packed into the panel.

EDIT: mentioning clipboard history


abb128:
Thanks! 120 seemed too small on landscape so I've changed it to 140

PR #1301, opened 2025-05-19
Fix crash and incorrect inline suggestion alignment

Adolar0042:
Having the `.horizontalScroll()` modifier on a LazyRow causes android to think it has an infinite maximum width, leading to a crash. 
Other than that the inline suggestions were getting cut off at the bottom due to the padding. As I couldn't correct the padding in a way that fixed the visual issues and the size of the elements didn't seem to depend on that padding I chose to vertically center them in the LazyRow instead.


Adolar0042:
Before:
image
After:
image

abb128:
Thanks, I think it's better to use a regular Row here actually because we're dealing with Views that can behave a bit abnormally, so I've switched it to use Row

PR #1319, opened 2025-06-01
Various changes to floating keyboard

Adolar0042:
I split my changes into 3 commits in case one of the alterations is unwanted.
Before padding fix:
image
After padding fix:
image

---

Before updated drag bar:
image
After updated drag bar:
image

---

Before top corner rounding:
image
image
After top corner rounding:
image
image


abb128:
Thanks!

PR #1383, opened 2025-07-01
Dead key support

tenextractor:
Added a `DeadKeyPreCombiner`, which considers any key with a combining character to be a dead key. This makes layouts with dead keys possible.


abb128:
Thanks!

PR #1422, opened 2025-07-15
Add keyboard preview on ThemeScreen

barmiro:
This PR adds a collapsible keyboard preview to the theme screen:
- hidden by default, with only a SmallFloatingActionButton visible;
- updates dynamically as the user changes the selected theme;
- works with API 30 & up, disabled on older versions (they don't handle insets consistently).

theme_preview

## Known issue
When the user clicks the back arrow in the screen title without first collapsing the keyboard, it takes a couple of frames for main screen elements to settle as the keyboard collapses.

bug

This doesn't occur with system back navigation (the keyboard collapses first). Fixing this issue would probably require changing how the ScreenTitle back button works, or using some workaround with navigation. Seems excessive for a one-off, especially given the low severity.


abb128:
Thanks!

PR #1427, opened 2025-07-17
Add Tibetan Extended Wylie combiner

tenextractor:
When paired with an appropriate layout, this would allow the input of Tibetan script using the Extended Wylie transliteration system. Uses the [ewts-converter](https://github.com/buda-base/ewts-converter) library (Apache-2.0 license).

https://github.com/user-attachments/assets/feaa2048-011c-4779-97f2-c4a357a48c56

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -351,8 +351,6 @@ dependencies {
     androidTestImplementation 'androidx.test:rules:1.5.0'
     androidTestImplementation 'androidx.test.uiautomator:uiautomator:2.2.0'
     androidTestImplementation 'androidx.annotation:annotation:1.0.0'
-
-    implementation 'io.bdrc.ewtsconverter:ewts-converter:1.6.0'
 }
 
 project.logger.lifecycle("versionCode = ${android.defaultConfig.versionCode}")
--- a/build.gradle
+++ b/build.gradle
@@ -351,6 +351,8 @@ dependencies {
     androidTestImplementation 'androidx.test:rules:1.5.0'
     androidTestImplementation 'androidx.test.uiautomator:uiautomator:2.2.0'
     androidTestImplementation 'androidx.annotation:annotation:1.0.0'
+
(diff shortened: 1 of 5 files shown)


tenextractor:
Layout added in https://github.com/futo-org/futo-keyboard-layouts/pull/202

build.gradle
```diff
@@ -351,6 +351,8 @@ dependencies {
     androidTestImplementation 'androidx.test:rules:1.5.0'
     androidTestImplementation 'androidx.test.uiautomator:uiautomator:2.2.0'
     androidTestImplementation 'androidx.annotation:annotation:1.0.0'
+
+    implementation 'io.bdrc.ewtsconverter:ewts-converter:1.6.0'
```
abb128, line 355: The library is only two Java files(?), I think I prefer the code to be vendored rather than adding a maven dependency

tenextractor: Done, please check the code

java/src/org/futo/inputmethod/event/combiners/wylie/EwtsConverter.java
```diff
@@ -0,0 +1,4 @@
+package org.futo.inputmethod.event.combiners.wylie;
+
+public class EwtsConverter {
+}
```
abb128, line 4: I think you might have left it blank accidentally

abb128: Also make sure you preserve their license notice since it's required for Apache https://github.com/buda-base/ewts-converter/blob/c1a6d3f6a5f32f32952f85afbe3610e852e96423/src/main/java/io/bdrc/ewtsconverter/EwtsConverter.java#L1-L19

tenextractor: Sorry, I committed all changes now

abb128:
Thanks!

Issue #821, opened 2024-11-14
Inline autofill bug


BabyOilJohnson:
### What happened?

More than telling, it's better to show. 
||
|-------|
| ![Screenshot_20241114-174912](https://github.com/user-attachments/assets/5d8cd6d4-06c2-4c69-8e0f-17aedfd33cd1) | 
As you can see when the inline autofill is appered scroll the autofill button to the left side gets overlaid on the more action button and on the other side there is an half pill sized thing for nothing. 

### Keyboard Version

0.1.24

abb128:
Inline autofill in Android is a bit strange. Android provides Views to render and it's made in a way that the keyboard cannot access the contents. They also render on top of everything. It's a bit difficult to work with and they cannot be clipped easily. As this is a visual bug it shouldn't really affect functionality other than having to unscroll it to toggle expanded actions

PR #1430, opened 2025-07-18
fix: inline suggestion chips not clipping horizontally

Adolar0042:
**NOTE: this is in no way a final fix for the issue as a whole, this only fixes the horizontal clipping**, the popups from keys getting pressed and everything else still gets ignored and rendered below the inline suggestions.
I've been holding back on committing this out of concern that the issue is forgotten or no longer prioritised and as a result not properly and fully fixed. So please, don't take this as a final fix, this should be nothing more than a "stopgap measure" of sorts.

An unfortunate side effect of this is losing the overscroll animation on the inline suggestions, as I've found no way to account for the extra padding and differing positioning during it.

## My limited understanding of the problem

_This is mostly conjecture on my end with no real proof attached to it._
image
When a inline suggestion view looks where/how it needs to render, it incorrectly assumes it is in view due to being entirely inside of the bounding box of the row it is in (I'm guessing it only looks at it's parent, not the parent's parent etc), completely ignoring the actual clipping bounds and anything that should render above it.

## How the fix works

I basically calculate the bounding box to clip to for each inline suggestion manually and tell each view to clip to it or fully make it invisible if the bounding box is empty/out of bounds. (Though as already communicated only for the horizontal bounds.)

## Before/After of the fix

https://github.com/user-attachments/assets/9869b14f-3efe-4bf2-83b5-34143327bcf0

### DIFF
--- a/java/src/org/futo/inputmethod/latin/uix/InlineSuggestionView.kt
+++ b/java/src/org/futo/inputmethod/latin/uix/InlineSuggestionView.kt
@@ -2,6 +2,7 @@ package org.futo.inputmethod.latin.uix
 
 import android.annotation.SuppressLint
 import android.content.Context
+import android.graphics.Rect
 import android.graphics.drawable.Icon
 import android.os.Build
 import android.util.Size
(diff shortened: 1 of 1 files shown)


BabyOilJohnson:
Will this PR fixes #821?

Adolar0042:
Partially, it fixes the horizontal clipping, but as stated in my pr, it doesn't fix stuff like popups of pressed keys.

abb128:
Thanks!

Issue #1417, opened 2025-07-12
Feature request: 2D Spacebar Swipe


AnalogAnalogue:
### Use Cases

When spacebar behavior is set to "swiping and long-pressing only moves cursor", the swipe input only registers movement along the x axis, and only allows moving the cursor forward and backward rather than up and down lines of text.

### Proposal

When spacebar behavior is set to "swiping and long-pressing only moves cursor", the keyboard should also register the vertical component of swipe inputs in addition to the horizontal input, allowing the user to move the cursor up and down lines of text similar to the functionality of the up and down arrows. This is the way Apple's stock iPhone keyboard works.

hassan-waris:
I just wanted to add this recording to show exactly what I would like to see from this ticket.

https://github.com/user-attachments/assets/2e6080f8-92a6-42a5-ba80-737c46dba20c

TheQwervy:
#260

PR #1443, opened 2025-07-22
Feature: Add multiplier to swipe-based cursor movement based on y

jones-josh:
Moving 1 character at a time makes the feature impractical outside of minor adjustments after tapping on text. This change splits the keyboard into 3 bands between the spacebar and the top of the keyboard, and when swiping, moves the cursor by 2^band characters.

Note that this would contrary to the open ticket #1417 as the y axis would be overloaded, perhaps a new setting to switch between behaviours would be desirable?

### DIFF
--- a/common/src/org/futo/inputmethod/latin/common/CoordinateUtils.java
+++ b/common/src/org/futo/inputmethod/latin/common/CoordinateUtils.java
@@ -91,4 +91,8 @@ public final class CoordinateUtils {
             @Nonnull final int[] coords) {
         setXYInArray(coordsArray, index, x(coords), y(coords));
     }
+
+    public static int clamp(int value, int min, int max) {
+        return Math.max(min, Math.min(max, value));
+    }
 }
--- a/java/src/org/futo/inputmethod/keyboard/PointerTracker.java
+++ b/java/src/org/futo/inputmethod/keyboard/PointerTracker.java
@@ -935,6 +935,10 @@ public final class PointerTracker implements PointerTrackerQueue.Element,
             }
 
             int steps = (x - mStartX) / pointerStep;
+            int STEP_MULTIPLIER_STAGES = 3;
+            int spaceY = oldKey.getY();
+            int step_multiplier = STEP_MULTIPLIER_STAGES - CoordinateUtils.clamp(y, 0, spaceY - 1) * STEP_MULTIPLIER_STAGES / spaceY - 1;
+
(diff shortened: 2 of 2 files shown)


jones-josh:
I will sign pending review.

Felix14-v2:
> text. This change splits the keyboard into 3 bands between the spacebar and the top of the keyboard, and when swiping, moves the cursor by 2^band characters.

Imo, that's impractical. If you want to increase the navigation speed, it would be much handy to multiply the cursor movement speed based on the swipe speed. Something like adaptive mouse pointer sensitivity on Windows.

Also there's another great improvement to the cursor movement, when moving your finger to the side of the screen activates automatic cursor movement:

https://github.com/user-attachments/assets/412b801b-88fd-4cdb-83c9-1176b8258cb5

But in this example, the cursor only moves one letter per tick at a fixed frequency, but in my opinion, it would be much better to move one word at a time, constantly increasing the speed at the same time.

jones-josh:
I had in mind how many media players work with seeking on touch devices. I disagree with it being based on movement speed out of purely personal preference but edge-based fixed rate word movement is a wonderful idea I much prefer to my solution.

Does that edge movement (with its current 1 char per step) have a ticket / PR? If so would be better to move to that and close this one but I can't find.

AnalogAnalogue:
I actually like your "bands" idea for cursor acceleration. Would it be possible to change the bands to work horizontally based on distance from the center of the spacebar? Like swiping space a bit left moves the cursor slowly backwards but swiping farther left moves the cursor more quickly? Same for swiping to the right. This would also leave the Y-axis free and could be implemented with the edge auto-scroll. Another option (and I have no idea if this is even possible) could be to bind the cursor positioning to the relative x component of the spacebar swipe as a related rate to the remaining space in the text line; like as a user swipes the spacebar to the left edge of the screen the cursor moves to the left edge of the line of text at a speed proportional to the distance between the start of the swipe and the edge of the screen. This would mean swiping the spacebar to full left screen edge always moves the cursor to the full left edge of the text line, and same for the right. I think Apple's stock iPhone keyboard works that way.

abb128:
IMO vertical cursor movement is requested far more often, and this would conflict with that unless we add more settings, which we're trying to avoid right now for fear of the keyboard becoming too complex, so I'll close this

jones-josh:
That's fair, I'll keep using my patch but understand the decision. If anyone requests the feature again do point them here please so they have a chance to do the same. Thanks for the keyboard!

IkXflow:
Yeah that feature would be amazing to have.

PR #1525, opened 2025-08-23
Add welsh emoji lookups

pjrobertson:
This change takes the existing `emoji_i18n.jsondlgz` and adds a `#cy` section at the end, with lookups in Welsh for emojis. Here are some screenshots of it in action:

| sick (sal) | bed (gwely) |
| ---| ---- |
|  emoji_sal |  emoji_gwely |

I'm attaching the translated json file for the Welsh, in case it's needed in the future for future builds.

[cy_emojis.json](https://github.com/user-attachments/files/21946628/cy_emojis.json)


abb128:
Where were the translations sourced from?

The jsondlgz is generated automatically from CLDR so it probably shouldn't be modified directly as it would make future updates more complicated. There was also already a #cy section in the json from the CLDR, is there a reason to replace it with a different one?

pjrobertson:
Thanks, abb128!

Many translations were missing for the original #cy section (e.g. some
emojis weren't searchable, and / or only searchable with a few keywords or
English. Also some important Welsh-context emojis were missing (e.g. Welsh
flag 🏴󠁧󠁢󠁷󠁬󠁳󠁿 - note this is also missing from #en).

It sounds like I should try and make those changes upstream in this case -
presumably you're pulling from here?
https://cldr.unicode.org/#translations-and-other-language-data

pjrobertson:
Oh, and the translations were created using a script with
google-cloud-translate, and then hand vetted by myself (sampled). Google
translate is very good for en<>cy due to the large corpus of en<>cy
available from the Welsh government (Welsh ML is something I work on also)

abb128:
I see, yes CLDR would be upstream. I'm not familiar with the CLDR contribution process but you can try contributing there (let me know if you have any luck). I can see about including the json as an override for Welsh in the meantime

pjrobertson:
Thanks, yeah I've just had a quick look at the unicode contribution process
and it seems... Complicated. Will let you know how I get on, but an
override might be useful initially!

PR #1570, opened 2025-09-10
Add more dictionaries by default

Shunnuo:
Added more dictionaries by default so more languages have spell check without downloading a dictionary.
It’s not for all languages, but it brings it on par with Heliboard and saves people time.


Shunnuo:
The languages added are from [android-keyboard/dictionaries/](https://github.com/futo-org/android-keyboard/tree/master/dictionaries)
They are included on the app, but not available to use, and still ask for download.

abb128:
There are no corresponding .dict files in the java/res/raw directory so this wouldn't compile. The files in the dictionaries dir are not shipped with the app. I'm not completely sure of their quality or how much size it would add to compile and include them

Shunnuo:
Can it be added for one language?

PR #1658, opened 2025-10-24
Make Armenian layout show the ASCII comma ',' on the comma key

tenextractor:
Fix https://github.com/futo-org/futo-keyboard-layouts/issues/28 . The last PR was unintentionally closed when I renamed the branch.


tenextractor:
The ARMENIAN COMMA is already inside the period key's `moreKeys`, so we can remove it from the comma key and show the ASCII comma there instead (as is done in Gboard).

abb128:
Fixed it in d996018b without realizing about this PR

PR #1660, opened 2025-10-24
Replace question mark with Persian equivalent

tenextractor:
Fix https://github.com/futo-org/futo-keyboard-layouts/issues/225


abb128:
Thanks!

PR #1805, opened 2026-01-04
feat: Add Quick Settings tile to temporarily hide keyboard

### LINKED ISSUE #1787: Feature request: Quick setting to hide keyboard.
### Use Cases

I read a lot of text files on my phone and when I want to highlight something, the keyboard pops up and the page scrolls up highlighting more than needed. I have to go back to remove the keyboard. (This happens a lot in note taking apps)

A **Quick settings tile** to disable the keyboard for some time will be an awesome feature. (Basically a one click toggle)

### Proposal

- This feature could be implemented through a quick settings tile in android. We would only need to scroll down from top and toggle this setting on and off. 

- Click the button to hide the keyboard (make it not pop up) and click it again to make it work normally. 
 
- I tried finding an alternative for the time being and the only keyboard I could find that supported it was by a developer that I do not recognize and the development of the app has stopped for quite some time.

### References

This is a very novel feature request if I am not wrong. I have looked through the issues.

nicomeemes:
Closes #1787

## Summary
Adds a Quick Settings tile (API 24+) that allows users to temporarily suppress the keyboard from appearing. When enabled, the keyboard will not pop up when text fields are focused, which is useful for reading/annotating text without keyboard interference.

## Changes

| File | Description |
|------|-------------|
| `java/src/org/futo/inputmethod/latin/qs/HideKeyboardTileService.kt` | New TileService that toggles keyboard suppression |
| `java/res/drawable/ic_keyboard_hide.xml` | Vector icon for the Quick Settings tile |
| `java/src/org/futo/inputmethod/latin/uix/Settings.kt` | Added `KEYBOARD_SUPPRESSED` preference key |
| `java/src/org/futo/inputmethod/latin/LatinIME.kt` | Modified `onShowInputRequested()` and `onEvaluateInputViewShown()` to honor suppress flag |
| `java/res/values/strings.xml` | Added tile labels and toast messages |
| `java/AndroidManifest.xml` | Registered the TileService |

## How it works
1. User adds the "Keyboard" tile from Quick Settings panel
2. Tapping the tile toggles suppression and shows a confirmation toast
3. When suppressed, the IME returns `false` from `onShowInputRequested()` and `onEvaluateInputViewShown()`, preventing automatic keyboard popup
4. Tap again to restore normal behavior

## Testing
- [ ] Add tile from Quick Settings panel
- [ ] Toggle ON → keyboard should not appear when tapping text fields
- [ ] Toggle OFF → keyboard appears normally
- [ ] Tile state persists across app restarts
- [ ] Toast messages display correctly

### DIFF
--- a/java/AndroidManifest.xml
+++ b/java/AndroidManifest.xml
@@ -63,6 +63,18 @@
         </service>
         -->
 
+        <!-- Quick Settings Tile for hiding keyboard (API 24+) -->
+        <service
+            android:name=".qs.HideKeyboardTileService"
+            android:icon="@drawable/ic_keyboard_hide"
+            android:label="@string/qs_tile_hide_keyboard_label"
+            android:permission="android.permission.BIND_QUICK_SETTINGS_TILE"
+            android:exported="true">
+            <intent-filter>
+                <action android:name="android.service.quicksettings.action.QS_TILE" />
+            </intent-filter>
+        </service>
+
         <!-- Activities -->
         <activity android:name=".uix.settings.SettingsActivity"
              android:theme="@style/Theme.AppCompat.DayNight.NoActionBar"
--- a/java/res/drawable/ic_keyboard_hide.xml
+++ b/java/res/drawable/ic_keyboard_hide.xml
@@ -1,0 +1,35 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+    android:width="24dp"
+    android:height="24dp"
+    android:viewportWidth="24"
+    android:viewportHeight="24">
+    <!-- Keyboard base -->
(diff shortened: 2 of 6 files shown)


abb128:
Thank you for the pull request but, I think my points in the linked issue still hold that this should be handled by a text editor app rather than adding a workaround to the keyboard. With this implementation it's likely going to be more common that a user will trigger it accidentally and not understand why the keyboard is no longer working, than actually gain any use from this functionality. Ultimately unless there's some other convincing use case I'd rather say no to this and ask you to request your text editor app to add a keyboardless viewing/highlighting mode, or use a text editor app that already has such a mode

Issue #707, opened 2024-06-29
Support for Traditional and Simplified Chinese Input


null-von-sushi:
I would like to request Chinese support, especially traditional mandarin.
To implement support for the Chinese group of languages you mostly just need Mandarin and Cantonese, with each having a normal and a simplified version.

Thinking about the layout of the FUTO Keyboard language settings, I imagine the following entries in _Language_
 - Cantonese (Simplified)
 - Cantonese (Traditional)
 - Mandarin (Simplified)
 - Mandarin (Traditional)
 
 The _Layout_ could then be used to select what type of IME to use. For Mandarin Hanyu Pinyin and Zhuyin input are important, while for Cantonese Sucheng, Cangjie and some sort of Cantonese Pinyin (there are different standards...) would be useful. For the initial implementation, I would start with Pinyin only. 
 
 It may be useful to use [RIME](https://rime.im) under the hood if possible as it is relatively multiplatform and supports many, many different layouts already and as such can easily handle Zhuyin, Pinyin as well as really cool but non-standard things (such as pinyin input with tones via the terra-pinyin schema). 
 
 
Some notes:
 - Traditional does not use European quotation marks. Its double quotation marks are `『...』` (U+300E LEFT WHITE CORNER BRACKET & U+300F RIGHT WHITE CORNER BRACKET respectively), while the single quotation marks are `「...」`  (U+300C LEFT CORNER BRACKET and U+300D RIGHT CORNER BRACKET respectively).
 - The enumeration comma (`、` U+3001 IDEOGRAPHIC COMMA) is used for listing things.
 - `《...》`,  `〈...〉`  and `【...】` are also often used, and it would be beneficial to have them as a long-press option on the quotation mark key, similar how long pressing `"` in the English layout gives options for `“ ...”` and ` «...»`

btyler:
Bump. I need to communicate in Chinese, and I don't want to have to switch to Gboard every time.
Is there any active development on this yet?

MattTheTekie:
I need to communicate in Chinese as well. Is there a way to contribute?

Timoite:
I have no idea why this feature request is in low priority

null-von-sushi:
Because it's complicated to implement *well* and because that's like no developers who actually care about this language to implement it I guess. It's a shame because if you only use European languages there's so many OSS keyboards, yet no that actually support all languages

btyler:
[...]

PR #1821, opened 2026-01-09
Chinese/Rime Engine Support

AstroNot233:
#707 Support Chinese/Rime

### Important
Rime is a schema-based Chinese Input Method Engine, which means without customize schema files, it cannot work. Surely I can copy some default schemas to assets and deploy them automatically. However, most of these schemas are licensed under LGPL, so I won't do so. Unfortunately if you never use Rime-based IMEs before, you need to learn where to place schemas you download.

### Note 
New dependency: [RimeJNI](https://github.com/AstroNot233/RimeJni.git)
After compilation, copy output to `libs/rime-release.aar`.
Later I will commit a layout for Chinese, it is completely the same as general QWERTY, but l created a directory to storage Chinese layouts for future implementation.


AstroNot233:
Sorry I forgot to configure the consumer proguard files for RimeJni. Please use the newest version of [RimeJni](https://github.com/AstroNot233/RimeJni.git)

TODO:
Usually, Chinese IMEs don't display preedit inline. I will add a bar over keyboard (the position of "more actions", is it its name?) to temp storage it.

abb128:
Can you link some of these schema files you expect to be used? Ideally we should make things work out of the box, but if it bloats the apk size too much for people who don't need it then it's better to make it a separate download (on Google Play we can do dynamic delivery based on device locales)

AstroNot233:
> Can you link some of these schema files you expect to be used? Ideally we should make things work out of the box, but if it bloats the apk size too much for people who don't need it then it's better to make it a separate download (on Google Play we can do dynamic delivery based on device locales)

Users can fetch these schemas from [here](https://github.com/rime/plum/tree/master?tab=readme-ov-file#packages)
Average 2kB I think, except for some dictionaries about 6MB.

AstroNot233:
> > Can you link some of these schema files you expect to be used? Ideally we should make things work out of the box, but if it bloats the apk size too much for people who don't need it then it's better to make it a separate download (on Google Play we can do dynamic delivery based on device locales)
> 
> Users can fetch these schemas from [here](https://github.com/rime/plum/tree/master?tab=readme-ov-file#packages) Average 2kB I think, except for some dictionaries about 6MB.

If you think it is acceptable, I can write a default schema deployer to copy those to `${storage}/Android/data/${package}/rime/shared`.
I have said LGPL may be a problem because I am not sure about FUTO's license

AstroNot233:
```
 Length Name                           
 ------ ----                           
        opencc                         // Simp<->Trad conversion, ~1MB. Apache2.0. Everything below is LGPL3.0
   5487 bopomofo.schema.yaml           // Mainly for Taiwan users
    653 bopomofo_express.schema.yaml   //
    801 bopomofo_tw.schema.yaml        //
 867883 cangjie5.dict.yaml             // Map cangjie5 code to characters (Not a dictionary)
   2535 cangjie5.schema.yaml           // Mainly for Hong Kong users
   2639 cangjie5_express.schema.yaml   // varient
6330573 essay.txt                      // Dictionary
 962079 luna_pinyin.dict.yaml          // Map pinyin to characters (Not a dictionary)
   2184 luna_pinyin.schema.yaml        // 
    556 luna_pinyin_fluency.schema.yaml// 
    775 luna_pinyin_simp.schema.yaml   // these 5 are varients
    655 luna_pinyin_tw.schema.yaml     // 
    420 luna_quanpin.schema.yaml       // 
   4965 pinyin.yaml                    // Spelling correction for luna_pinyin
   2212 punctuation.yaml               // For Chinese styled punctuations
2330785 stroke.dict.yaml               // Map stroke code to char
   3156 stroke.schema.yaml             // To be honestly nobody use it. Sometimes as an option for some customized schema to provide strokes query when users forget how to write that char (XD)
1898902 terra_pinyin.dict.yaml         // 
   4713 terra_pinyin.schema.yaml       // Another pinyin implementation, less popular than luna
   1178 zhuyin.yaml                    // Mainly for Taiwan users
```

Only luna-pinyin and its derivatives are support now, because others need specific keyboard layouts.

abb128:
I think due to the size of the dictionary we will end up needing to make it a separate download.

I've been testing it out and it seems to be working nicely, however I can't seem to get it to type Simplified Chinese even when toggling it in rime dashboard or selecting luna_pinyin_simp, it's always typing Traditional Chinese characters. Is it working on your end and I'm missing something or is it just broken?

Last thing, librime adds native .so files which total 43.4MiB uncompressed, 11.5MiB compressed. Do you reckon there's any way to get that number down? Modern Android advice is to not compress native libraries, so that they can be memmapped from the APK directly, so currently this PR increases app size from 130MiB to 176MiB. Maybe we can enable compression but I'd be worried about whether something would break in the wild because of it

AstroNot233:
@abb128 Thank you for your test!

## Simplified and Traditional
For simp and trad, it may be confusing because RIME has a convertor itself. So please directly switch the 4th button "傳" to "简", which is traditional form of "trad-" to simplified form of "simp-". If it doesn't work, please ensure `${ExternalStorage}\rime\shared` contains OpenCC. 
Considering that if some one use Chinese IME, they are expected to be able to understand these two characters, I think it makes sense using these two characters as icons. 

## About Compression
I enabled compression then built alpha release ver of FUTO and have used it as my main IME for 10 days already. I haven't met any problem so far, so maybe it doesn't matter to enable compression I think.
By the way, is it possible to split libs for 4 ABIs? For v8a, all libs(rime included) occupy only 21MiB. Even before rime being added, it can now reduce 24MiB immediately.

## Another Way
Since other languages don't have default dictionaries, how about just regarding all default schemas as a dependency? Then replace download page of Chinese dictionary with schemas and dictionary collection zip. I will add an entrance like Japanese in Settings to directly import the zip file. It's the most balanced solution I can give. If you think it is OK, I will do that.

AstroNot233:
Ah, actually rime has simp->trad and trad->simp 2 convertors and they are separated options.
I think it doesn't make sense if you turn on both, so I only designed one button to turn one on and the other off, even if I knew turning off both means no conversion. 
My opinion: If someone wants to type simplified, it's obviously more natural to tap a button than switching to another schema.
From this perspective, there is no use to keep simp and trad variants. 🤔Anyway, default schemas should stay minimized.

AstroNot233:
Sorry, I met some difficulties when trying adding zip importer.

There should be a general unzip module, but I did `grep` and found features like importing settings or themes make wheals separately. It will be weird if I try quirky ways to pass necessary variables, like context, to the settings button and write unzip function again. On the other hand, this PR is for Chinese support. Refactoring these utils is not my purpose.

So sadly, this PR can only end up with that users need to download zip then unzip to `${ExteranlStorage}/rime/shared` by themselves.

abb128:
> If it doesn't work, please ensure ${ExternalStorage}\rime\shared contains OpenCC.

I think that's my problem, wasn't aware opencc is necessary


> So sadly, this PR can only end up with that users need to download zip then unzip to ${ExteranlStorage}/rime/shared by themselves.

No worries, I can implement an importing feature. Thanks for your help.

Do you want me to implement the floating preedit as well? I can base it off of how Fcitx5 looks on Android. Also curious your thoughts about the expandable suggestion bar, the current implementation is intended for Japanese, should it be changed too?

AstroNot233:
> Do you want me to implement the floating preedit as well?

I am not sure about if you mean:



Preedit isn't displayed inline but in ime. That's great and I may support move cursor of preedit. I had given up because it's hard to judge whether the user wants to move insert cursor or composing cursor, but this will make it clearer and no need to calc offset. 

---

> Also curious your thoughts about the expandable suggestion bar, the current implementation is intended for Japanese, should it be changed too?

Sorry, I don't understand what "be changed" mean. Did you mean it will be refactored? Or it is intended for Japanese  only, so Chinese need another one?

I think current expandable suggestion bar  is just "useable" but not pretty, to be honest.



As you can see, I don't know why the first candidate is much longer. It looks a little ugly. However, it's unnecessary to have a new one, so just using existing one is okay.

abb128:
Yes, the preedit hovering above the keyboard rather than being typed inside the text field until it's committed is what I mean.

> Did you mean it will be refactored? Or it is intended for Japanese only, so Chinese need another one?

When I test Chinese in Fcitx5 or other keyboards, the expanded view is different. There's up/down arrow keys, backspace and enter on the right side. The current implementation doesn't have this, it only shows the candidates and you need to collapse it to access backspace or enter. As I understand the backspace functions like undo if you picked the wrong character, but I'm not sure how enter is used in typical day-to-day typing
image

AstroNot233:
> The current implementation doesn't have this, it only shows the candidates and you need to collapse it to access backspace or enter.

Enter is to commit instantly. It is mainly used when typing English in Chinese mode. However when in the candidate panel, it means "the user typed a long string, then picked some candidates, and finally the string end with an English word, so the user pressed enter to commit". That's really rare. 

Backspace is the only necessity. Adding scroll up and down and enter makes it look prettier, and more like mainstream IMEs. 

Thanks for your commitment!

abb128:
@AstroNot233 Is there any reason for requiring Lua in RimeJNI? It doesn't look like it's being used under the standard configuration. If we don't include it, the size can be reduced a little bit by not needing to unnecessarily include Lua VM

AstroNot233:
Not a must.
I just saw [Trime](https://github.com/osfans/trime) support it, and librime-lua is almost an officially recognized extension, so I add it.
Some(few) schemas use lua, like typing `/date` and suggesting `22/01/2025`, typing `=12*3+4`and suggesting `12*3+4=40`, etc. For these, disabling lua won't lead to crash. General language services will still work.

abb128:
@AstroNot233 I'm considering removing the RIME Dashboard action and instead add two languages that would automatically toggle simplification: Chinese (Simplified) and Chinese (Traditional). Schema switching could be in the settings menu (I'm assuming people won't be switching schemas super often) among other settings, and the schema would default to luna-pinyin in a new installation. I'm not entirely sure what other settings should be in the settings menu. What are your thoughts?

AstroNot233:
@abb128 
Most Chinese IMEs(not Rime, but they all have these features) make these buttons easy to access. Features like temporarily full/half -width convert and traditional/ 
simplified convert are often used actually. In Rime distros for PC, just `Ctrl+accent` to call a menu which shows the schema list and other variables. That's why I place it in the action instead of settings. Splitting all these features and placing them in the action directly might be better, but that will make the action list over-crowded.

Default schemas or default variables like `traditional` can be appointed in `default.yaml`. Users can override it by `user/default.custom.yaml`.

AstroNot233:
> I'm assuming people won't be switching schemas super often

I am a shape-based input method user. Sometimes we may forget how to write that character, so quick switching to pinyin then switching back is common.
For phonetic-based input method users, they sometimes see a character but don't know how to pronounce, they usually switch to hand-writing(not supported by Rime yet, but Rime has strokes mod) then switch back.

abb128:
Traditional/simplified quick switching would still be possible by adding both languages and using the language switch key

What does half width vs full width do exactly? I couldn't find such an option in Gboard or SwiftKey, are they just sticking to one?

AstroNot233:
@abb128 
> Traditional/simplified quick switching would still be possible by adding both languages and using the language switch key

This can be nice, but all Chinese IMEs place that button here.

(Screenshot from my friend, simp-trad convert at R2C4)
How about doing what you suggest but keeping the button?

---

> What does half width vs full width do exactly?

This is the same as Japanese. Half is default one. When switch to full, typing `Hello, world!` to get `Ｈｅｌｌｏ，　ｗｏｒｌｄ！`
**Note**: This is irrelevant to `isAsciiPunt`. If `isAsciiPunct && fullShape`, `.^$` ->`．＾＄`, instead of `。……￥`. However, no matter whether `fullShape`, if `! isAsciiPunct`, `.^$` always -> `。……¥`. 

P. S. : I agree `isAsciiPunct`and `fullShape`are useless, but removing them from dashboard will make dashboard look ugly. How do you think?

abb128:
I could maybe add an action for toggling simplification so that the dashboard is not necessary but the action system will need some rework because it doesn't support toggle actions currently.

RE: handwriting and stroke input, I think the ideal solution is to have different layouts for these anyway. Here is an image of how SwiftKey does it
Screenshot_20260123-155019
So again, switching between stroke and pinyin would be handled by the language switching menu instead of needing an inline schema switcher. I think it makes more sense this way anyway, and prevents average users from messing it up and selecting the wrong schema. Handwriting is a much more complex issue that we're not going to focus on for now

AstroNot233:
@abb128 

> I think it makes more sense this way anyway, and prevents average users from messing it up and selecting the wrong schema. 

Corresponding particular schemas to particular layouts, and displaying schema in language list, are two of my todo, but I don't have any idea yet. That's great if you can solve that!

---

> I could maybe add an action for toggling simplification so that the dashboard is not necessary

Yes, after splitting schemas and simp-trad convert, others can be moved to settings like Japanese, and dashboard is no longer needed. 

---

Here comes another problem(a potential bug):
Toggling full-shape or others needs Rime instance being alive. I choose to create an instance in ChineseIME and borrow it to actions. Dashboard needs Chinese selected, so it's a natural insurance.
But for settings, it's complicated. It's possible that a user downloads FUTO then directly goes to Chinese settings to import default schema pack and configures some settings. Before switching to Chinese, no instance is created. So the backdoor `fun requestTakeOver(who: Any)` I left in ChineseIME will no longer help. You need to find another way, and ensure Rime is singleton. 

From this aspect, that's why I have tried to persuade you leave dashboard there. I think that's too difficult and not worth such hard refactoring.

abb128:
> It's possible that a user downloads FUTO then directly goes to Chinese settings to import default schema pack and configures some settings

I think it's not that complicated. The settings can be stored as standard datastore settings, and ChineseIME reads and applies them on input start. This is similar to how JapaneseIME works

abb128:
Btw, I have merged the changes into the [zh branch](https://github.com/futo-org/android-keyboard/tree/zh) and will continue development there. Nightly builds are available at https://dl.keyboard.futo.org/nightly-zh.apk

AstroNot233:
@abb128 

I tried using this nightly for one day. Here is my feedback: 

### defualt.yaml update
``` yaml
# default.yaml
# ...

schema_list:
  - schema: luna_pinyin
  - schema: luna_pinyin_simp
  - schema: luna_pinyin_fluency
# Is this missing?
  - schema: luna_pinyin_tw
# Redundant:
#  - schema: bopomofo
#  - schema: bopomofo_tw
#  - schema: cangjie5
#  - schema: stroke
#  - schema: terra_pinyin

# ...
```

---

### redundant files
* `key_bindings.yaml`
  It's for PC hotkeys.
* `symbols.yaml`
  For punctuations, there's already `punctuation.yaml`. For symbols, long press is better than `/xx`.

---

### import dict
I know current nightly hasn't implemented import yet. Should .dict be extracted after import instantly? Or lazily check its timestamp when users trigger redeploy?
By the way, after importing, warning that `No dictionary installed...` still shows every time.
*Update*: This is because import for `zh_HK_#Hant`, dict will be renamed to `dictionary_zh_HK_HHant.dict`instead of `dictionary_zh.dict`. Also hacked by `zh_TW_#Hant`. Maybe just match `.startsWith("zh")` and distinguish traditional or simplified by `.endsWith("#Hant")` or `.endsWith("#Hans")`? More directly, changing the behavior of auto adding languages after installation? 

---

### about copy
I am not sure if it is my fault:
When on Chinese, copy something won't show a suggestion but paste(or "send" in some apps) will. Even worse, I don't know how to cancel it, and new suggestions will be overrode by it. I didn't see JapaneseIME have relevant codes and it seems to have the same problem. If it is a general bug, I'll open an issue.

wjouqd:
I have the same issue with copy, no issue with "Import dict".

AstroNot233:
> I have the same issue with copy, no issue with "Import dict". 

@wjouqd Are you using Simplified Chinese as system default language? I checked the commit history, @abb128 wrote logic of forcing language to be "zh" for `zh_CN_#Hans` but not for HK and TW. Manually added Chinese is "zh", with expected behavior.

abb128:
There should be no issues with importing the dictionary on the latest nightly. It was broken in the first build. Make sure you're updated and let me know if there are further issues.

Clipboard visibility is handled by helper.updateUiInputState, I'll add it

wjouqd:
> > I have the same issue with copy, no issue with "Import dict". 
> 
> @wjouqd Are you using Simplified Chinese as system default language? I checked the commit history, @abb128 wrote logic of forcing language to be "zh" for `zh_CN_#Hans` but not for HK and TW. Manually added Chinese is "zh", with expected behavior. 
> 

No, I am using English as system default language, and Simplified Chinese in input method.

wjouqd:
> > I have the same issue with copy, no issue with "Import dict". 
> 
> @wjouqd Are you using Simplified Chinese as system default language? I checked the commit history, @abb128 wrote logic of forcing language to be "zh" for `zh_CN_#Hans` but not for HK and TW. Manually added Chinese is "zh", with expected behavior. 
> 

No, I am using English as system default language, and Simplified Chinese in input method.

wjouqd:
I encountered another glitch I am not sure if it's an issue: every time switching back to Futo, the default Chinese input mode is "traditional Chinese".  I have to change it into simple Chinese manually.

AstroNot233:
> @wjouqd : I encountered another glitch I am not sure if it's an issue: every time switching back to Futo, the default Chinese input mode is "traditional Chinese". I have to change it into simple Chinese manually.

Rime has two convertors, simp->trad and trad->simp. Both are off default. Toggle that button will switch to one on and one off. Luna pinyin outputs traditional Chinese because trad->simp is more accurate(eg: simplified Chinese merges `干幹乾` to `干`).

wjouqd:
Thanks a lot for your explanation about the simp->trad and trad->simp converters. 

> > @wjouqd : I encountered another glitch I am not sure if it's an issue: every time switching back to Futo, the default Chinese input mode is "traditional Chinese". I have to change it into simple Chinese manually.
> 
> Rime has two convertors, simp->trad and trad->simp. Both are off default. Toggle that button will switch to one on and one off. Luna pinyin outputs traditional Chinese because trad->simp is more accurate(eg: simplified Chinese merges `干幹乾` to `干`).

abb128:
Stroke input layout is added in latest nightly (it requires dictionary update to v26.01-1, since v26.01 didn't include stroke schema). I am not entirely sure what the keys correspond to in RIME other than the basic five strokes, so I have left a bunch of space keys in the 2nd and 3rd rows for now

AstroNot233:
> I tried using this nightly for one day. Here is my feedback:
> 
> ### defualt.yaml update
> ```yaml
> # default.yaml
> # ...
> 
> schema_list:
>   - schema: luna_pinyin
>   - schema: luna_pinyin_simp
>   - schema: luna_pinyin_fluency
> # Is this missing?
>   - schema: luna_pinyin_tw
> # Redundant:
> #  - schema: bopomofo
> #  - schema: bopomofo_tw
> #  - schema: cangjie5
> #  - schema: stroke
> #  - schema: terra_pinyin
> 
> # ...
> ```
> 
> ### redundant files
> * `key_bindings.yaml`
>   It's for PC hotkeys.
> * `symbols.yaml`
>   For punctuations, there's already `punctuation.yaml`. For symbols, long press is better than `/xx`.

@abb128 Did you forget to change `default.yaml` and remove `key_bindings.yaml` and `symbols.yaml`?
The `schema_list` in `default.yaml` need to be cleaned, because no file matches bopomofo, etc. now. I don't know whether selecting a schema which doesn't exist will cause a bug or not.

AstroNot233:
> Did you forget to change default.yaml and remove key_bindings.yaml and symbols.yaml?
The schema_list in default.yaml need to be cleaned, because no file matches bopomofo, etc. now. I don't know whether selecting a schema which doesn't exist will cause a bug or not.

@abb128 I opened a new repository [futo-zh-zidian](https://github.com/AstroNot233/futo-zh-zidian.git). After cloning, you can use `./export.sh` or `./export.sh "v26-01"` to generate the .dict file at `dict/`.

ResurFace382500000:
@AstroNot233 

How to install this on futo-keyboard?

abb128:
@ResurFace382500000 Please use the following nightly branch: https://dl.keyboard.futo.org/nightly-zh.apk

It will be merged into mainline nightly soon

ResurFace382500000:
> @ResurFace382500000 Please use the following nightly branch: https://dl.keyboard.futo.org/nightly-zh.apk
> 
> It will be merged into mainline nightly soon

Thanks! Great work!

wjouqd:
@AstroNot233 
The punctuation for Chinese input seems to be the same to English input method in the latest version-11625.

AstroNot233:
> @AstroNot233 
> The punctuation for Chinese input seems to be the same to English input method. 

@wjouqd Sorry that I changed the strategy of handling punctuations:
- Previous:
  - Always send ',' '.' etc. to Rime. Then let Rime decide to output '，' or ','.
  - Disadvantage: this convert can happen accidentally. E. g. when you want to type '$', Rime will still "cleverly" turn it into '￥', '_' will be forced convert to "——" and '^' will be "……".
- My solution:
  - Modify the layout of FUTO to let user directly input '，' or ',' in case they want. Since symbol input is much more convenient on mobile platforms than on PC, there is no need to apply previous method which is for PC. 
  - However, @abb128 didn't merge that change on layout. In that PR, we have some disagreements about how I treat floating preedit bar, so he didn't merge anything, including the layout. Probably it's because he didn't notice that. 

For me, a new semester began. I need several weeks to adjust myself, I cannot return coding very soon. So sorry :(

If you have some suggestions, we are happy to receive your comments or even PR on the "zh" branch.

AstroNot233:
My current layout looks like this. Each punctuations have half and full width choices, with full width as default(long press and release directly).

However, I cannot find a way to remove " . " key, as it seems to be added automatically...

wjouqd:
Thanks a lot for your explaination and I understood it is not a bug, but a temporary solution. It is not so urgent, let's take time and try to find a good solution. 

> > @AstroNot233 
> > The punctuation for Chinese input seems to be the same to English input method. 
> 
> @wjouqd Sorry that I changed the strategy of handling punctuations:
> - Previous:
>   - Always send ',' '.' etc. to Rime. Then let Rime decide to output '，' or ','.
>   - Disadvantage: this convert can happen accidentally. E. g. when you want to type '$', Rime will still "cleverly" turn it into '￥', '_' will be forced convert to "——" and '^' will be "……".
> - My solution:
>   - Modify the layout of FUTO to let user directly input '，' or ',' in case they want. Since symbol input is much more convenient on mobile platforms than on PC, there is no need to apply previous method which is for PC. 
>   - However, @abb128 didn't merge that change on layout. In that PR, we have some disagreements about how I treat floating preedit bar, so he didn't merge anything, including the layout. Probably it's because he didn't notice that. 
> 
> For me, a new semester began. I need several weeks to adjust myself, I cannot return coding very soon. So sorry :(
> 
> If you have some suggestions, we are happy to receive your comments or even PR on the "zh" branch.

PR #1833, opened 2026-01-16
Add URL tracking parameter stripping for clipboard

### LINKED ISSUE #556: Remove Tracking Parameters from URLs
### Use Cases

Basically a feature which you can toggle to remove parameters such as ?shareid/si , ?feature=share etc. from urls copied to clipboard. This can be performed either when url is copied or when pasted in a text field. This would be great for privacy of users and prevent a lot of tracking.

### Proposal

Extensions such as Ublock and ClearURL already do this in browsers with help of blocklists. Maybe you can create a blocklist which will automatically remove the mentioned parameters from copied text.

### References

_No response_

ftlframe:
Closes #556 - automatically removes tracking parameters
(utm_*, fbclid, si, etc.) from URLs copied to the clipboard using the ClearURLs ruleset.
- Add UrlCleaner.kt utility that parses url-rules.json and strips tracking params
- Add ClipboardStripUrlTracking setting (disabled by default)
- Modify ClipboardHistoryManager to clean URLs when copied
- Modify QuickClip to display cleaned URLs in suggestion bar
- Add url-rules.json from ClearURLs project to assets
   
**Testing:**
- "Remove URL tracking parameters" in Clipboard settings
- URLs with tracking params (YouTube, Amazon, etc.)
- Verify tracking params are stripped when pasted

### DIFF
--- a/java/assets/url-rules.json
+++ b/java/assets/url-rules.json
@@ -1,0 +1,3170 @@
+{
+    "providers": {
+        "amazon": {
+            "urlPattern": "^https?:\\/\\/(?:[a-z0-9-]+\\.)*?amazon(?:\\.[a-z]{2,}){1,}",
+            "completeProvider": false,
+            "rules": [
+                "p[fd]_rd_[a-z]*",
+                "qid",
+                "srs?",
+                "__mk_[a-z]{1,3}_[a-z]{1,3}",
+                "spIA",
+                "ms3_c",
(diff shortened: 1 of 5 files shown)


abb128 (changes requested):
Thanks for the PR. I haven't tested the changes yet but a couple things stood out to me in the code

SamCyanide:
Would love to see this implemented

PR #1857, opened 2026-01-31
Revert ü convertor; implement Chinese qwerty layout; optimize punctuation input.

AstroNot233:
@abb128

* Revert [ü convertor](https://github.com/futo-org/android-keyboard/commit/2d5841dfa7bb631cfd6fe0ab26e370607ac893b8#commitcomment-176076323)
* Implement [Chinese qwerty layout](https://github.com/futo-org/futo-keyboard-layouts/pull/263)
* Optimize punctuation input
  Modified schema collection is at [futo-zh-zidian](https://github.com/AstroNot233/futo-zh-zidian). Clone and use `./export.sh [version_name]` to get .dict file.

Btw, I have a question:
https://github.com/futo-org/android-keyboard/blob/2d5841dfa7bb631cfd6fe0ab26e370607ac893b8/java/src/org/futo/inputmethod/engine/general/ChineseIME.kt#L208
What does this actually do? I am confused by clearing it and inputting it again. And  on my device, it seems to be buggy: after clicking preedit bar, every input behavior will be super strange.

### DIFF
--- a/java/res/values/strings-uix.xml
+++ b/java/res/values/strings-uix.xml
@@ -290,6 +290,8 @@
     <string name="settings_search_enter_your_search">Enter a search term…</string>
     <string name="settings_search_no_options_found">No results found</string>
     <string name="misc_settings_title">Miscellaneous</string>
+    <string name="setting_section_collapsed">Collapsed</string>
+    <string name="setting_section_expanded">Expanded</string>
 
     <!-- When adding a new language -->
     <string name="language_settings_title">Languages &amp; Models</string>
--- a/java/src/org/futo/inputmethod/latin/uix/settings/Components.kt
+++ b/java/src/org/futo/inputmethod/latin/uix/settings/Components.kt
@@ -808,16 +808,20 @@ fun<T> DropDownPicker(
 
 @Composable
 fun CollapsibleSection(title: String, modifier: Modifier = Modifier, section: @Composable ColumnScope.() -> Unit) {
+    val context = LocalContext.current
     var expanded by remember { mutableStateOf(false) }
-
 
     Column {
         Row(
             Modifier.fillMaxWidth().heightIn(min = 44.dp).clickable {
                 expanded = !expanded
             }.padding(16.dp).semantics {
-                // TODO: Localization
-                stateDescription = if(expanded) "Expanded" else "Collapsed"
+                stateDescription = context.getString(
+                    if(expanded)
+                        R.string.setting_section_expanded
+                    else
+                        R.string.setting_section_collapsed
+                )
                 role = Role.DropdownList
             }
         ) {
@@ -852,8 +856,8 @@ fun<T> DropDownPicker(
     scrollableOptions: Boolean = false,
     modifier: Modifier = Modifier
 ) {
+    val context = LocalContext.current
     var expanded by remember { mutableStateOf(false) }
-
 
     SpacedColumn(4.dp, modifier = modifier.semantics {
         role = Role.DropdownList
@@ -868,8 +872,12 @@ fun<T> DropDownPicker(
             ).heightIn(min = 44.dp).clip(DropDownShape).clickable {
                 expanded = !expanded
             }.padding(16.dp).semantics {
-                // TODO: Localization
-                stateDescription = if(expanded) "Expanded" else "Collapsed"
+                stateDescription = context.getString(
+                    if(expanded)
+                        R.string.setting_section_expanded
+                    else
+                        R.string.setting_section_collapsed
+                )
                 role = Role.DropdownList
             }
         ) {


AstroNot233:
@abb128 Hello! Do you mind that I rewrite floating preedit preview? Especially editing, which I think is not that useful: we can learn from Japanese, using arrow keys to move cursor, in order to avoid complicated preedit preview processing.

Rime doesn't support "move cursor to x", but allows "move cursor by x". This is because when cursor is at, like "你|hao", then tap DPAD_LEFT, it will automatically undo the selection of "ni", and preedit will be "ni|hao". Do once again, some schemas get "|nihao", some get "n|ihao". This makes it hard to process when user select a word, then tap the middle of that word. So the best practice is, only displaying caret positions(I added getting caretPos in RimeJNI), allowing LR moving instead of clicking to move.

abb128:
@AstroNot233 Samsung and iOS support tapping to move and it seems more convenient so I'd prefer tapping to be supported. Haven't seen any that use arrows, it's more of a Japanese thing. Plus Japanese also supports tap to move

AstroNot233:
> @AstroNot233 Samsung and iOS support tapping to move and it seems more convenient so I'd prefer tapping to be supported. Haven't seen any that use arrows, it's more of a Japanese thing. Plus Japanese also supports tap to move

@abb128 I downloaded Trime and Fcitx5. Trime supports clicking to move cursor but has very low precision; Fcitx5 doesn't allow that at all.

Due to that Rime itself doesn't allow to set caret position, I can only use while scope to send XK_Left or XK_Right. Here is the complexity:
Take `你好shi jie|` as pre-edit for example. (the space between `shi` and `jie` is auto inserted by luna-pinyin)
The tuple (x, y, z) after `preedit` is for (caretPos, selStart, selEnd), which is provided by Rime. `selStart` and `selEnd` mean candidates are given based on this closed interval.
* (Origin) `你好shi jie` caret=9; sel=[2,9]
* XK_Left -> `你好shi|jie` caret=5; sel=[2,5]
  No space, and auto skip a full word. Candidates are given based on `shi`.
* XK_Left -> `你好|shi jie` caret=2; sel=[2,9]
  When `caretPos == selStart`, `selEnd == preedit.length`.
* XK_Left -> `ni hao|shijie` caret=6; sel=[0,6]
  Undo selection, now `selStart` falls back to `0`, so `selEnd` is set to `caretPos`.
* XK_Left -> `ni|haoshijie` caret=2; sel=[0,2]
* XK_Right -> `ni h|aoshijie` caret=4; sel=[0,4]
  Only move one letter, candidates match `ni h*`

This makes it very hard to calculating now many left shifts is needed when a user click the # marked position in `ni h#ao shi jie|`. What's more, notice that `fa# lü|` is expected to be `fa|lv` after moving. This hacks using detecting substring as while scope condition. For other schemas, what if some expand `v` to `zh`, some compact `ng` to `ŋ`?

You don't need to fully understand them, since you cannot change this behavior after all. Rime is initially designed for PC, so only uses Left and Right to move caret. Trime seems to try estimating but very inaccurate, sometimes left-shift can be correct, but never for right-shift.

So maybe rather than estimating, directly sending `shift+Tab` to Rime (Left shift by words) when tapping floating preedit, sending `Tab` to Rime (Right by words instead of letters) when long press can be prettier and much easier to implement.

AstroNot233:
@abb128 I redesigned floating pre-edit bar (Sorry to remove some of your codes). Now its onClick behavior can be provided by callback function. For Rime, now a single tap will send `XK_Left`, and long press will send `XK_End`.

abb128:
IMO the updated preedit is still worse for usability compared to the original, and requires the user to manually tap several times to go back and still need to backspace some to edit the middle of a word. It does feel hacky to clear the composition and retype it, and needs more work to function properly with different schemas, but is better from a usability standpoint. I would rather write special cases to deal with the schemas we'll support, and possibly have some things break for the 1% of people that might use custom schemas, rather than compromise on the UX for the 99%

AstroNot233:
@abb128 Can you describe how to calculate how many Lsh and Rsh are needed? I'll check Rime's source code to see if it is possible to implement, and if possible it's better to encapsulate the logic at RimeJNI.

Plus, the API header of Rime says, simulate_key_sequence is only intentioned to be used for debugging, instead of doing that.

```c++
  // testing

  Bool (*simulate_key_sequence)(RimeSessionId session_id,
                                const char* key_sequence);
```

abb128:
I could add a hook for when the cursor position changes in the field, and then subtract that from the current RIME position. However, I want to support editing middle of words as well, e.g. if the user types "zhingguo" and wants to correct the i to an o by tapping there, backspacing once, and typing an o. It might be a big task to make RIME support this

abb128:
Btw I see you added longpressable symbols but, is typing ü by longpressing v or u supposed to work? When I try it, it doesn't seem to work
<video src="https://github.com/user-attachments/assets/878d96ce-bcd4-48d5-afd2-2493f25fc92a">

AstroNot233:
@abb128 
> I could add a hook for when the cursor position changes in the field, and then subtract that from the current RIME position. However, I want to support editing middle of words as well, e.g. if the user types "zhingguo" and wants to correct the i to an o by tapping there, backspacing once, and typing an o. It might be a big task to make RIME support this

Okay, I tracked the call stack of process_key, and found `left` *IS* to Lsh a whole spelling unit (or "word", but spelling unit is irrelevant to space`' '`). So we need to calculate both Lsh and Rsh.

> Btw I see you added longpressable symbols but, is typing ü by longpressing v or u supposed to work? When I try it, it doesn't seem to work

These alphabets with accents are for pinyin input shortcuts, not for typing characters.

AstroNot233:
Oh I have an idea, before committing, we also maintain a string recording the raw input of user.(convenient and robust for all schemas) When the user click the floating bar, display the raw input string and the user directly edit that(no keys sent to Rime). After editing, clear_composition and simulate_key_sequence(maybe doesn't matter?).

The only shortcoming is that, if the user select some candidates, selecting behaviors will not be recorded.

abb128:
> These alphabets with accents are for pinyin input shortcuts, not for typing characters.

What do you mean? Can you give an example of when these characters would be used in the keyboard?

AstroNot233:
> What do you mean? Can you give an example of when these characters would be used in the keyboard?

Tom: "那个字读zhōng" (That character should be read as zhōng)

Just the same as symbols like @#$% etc.

abb128:
> Oh I have an idea, before committing, we also maintain a string recording the raw input of user.(convenient and robust for all schemas) When the user click the floating bar, display the raw input string and the user directly edit that(no keys sent to Rime). After editing, clear_composition and simulate_key_sequence(maybe doesn't matter?).

This does seem to work, but there is a downside that this doesn't include spaces between words for easier visual separation. Although I'm also wondering if the spaces may also be misleading in the end because they are basically ignored, and attempting to fix a typo that has merged or split a word by attempting to remove or insert a space would be futile

abb128:
I implemented editing based on keeping track of raw input. This works more robustly now even supporting double pinyin

AstroNot233:
@abb128 
I downloaded the nightly build. Here are some other feedbacks:
1. Was it a bug that traditional and simplified convertor didn't work?
2. I didn't find where to change the schema.
3. I didn't know how to (before I check source code) force re-deploying after importing user schemas.
4. Actually fuzzy-pinyin behaves very bad. So most of the time those schema authors don't suggest turning on that.
  e.g. If a char coded "xu" has a higher weigh than a char coded "zi", then even if you typed "zi", 1st candidate will still be "xu".
5. For some shape based schemas, like wubi(the most popular), which are "length-fixed" coded(for wubi, max code length is 4). That means if I type "abcde", the fifth letter 'e' will auto commit the first candidate, then the preedit will be 'e'. Current floating pre-edit bar didn't expected that. I checked current implement, rawInput should be cleared when received commit msg or onEndInput, etc., instead of current situation.

I'll open another PR after fixing some of them.

abb128:
The intended way is to use the dictionary from here, v26.02: https://keyboard.futo.org/dictionaries?locale=zh

1. It should work, by default it's based on the added locale's script or country tag, but there is an option in settings to force it on or off
2. Schema switching is handled automatically based on layout and the Chinese settings where you can select Double Pinyin
3. It should be redeployed automatically when you import the zidian.dict file and open the keyboard once. Generally I don't think most users should have to know about the concept of redeploying etc, things should just work automatically
4. That's unfortunate, but it's turned off by default anyway and other keyboards seem to have it. Although I'm not sure I understand your example, there is no x = z or u = i. My testing has been to enable z = zh and type "zong guo" and see if zhong guo comes up
5. I will look into wubi. If there are any other schemas you think should be supported by default other than luna-pinyin, double-pinyin and stroke let me know

AstroNot233:
Oh it looks like I used the wrong word. Maybe not fuzzy pinyin, I mean correction? It's for correcting typo, but it works really bad lol.

abb128:
You can turn off the correction now by turning off Autocorrect and Show correction suggestions in settings, but from my testing it's way worse without it. You can mistap one letter and it completely ruins the word and maybe surrounding words. So I think it's better with it enabled

AstroNot233:
> You can turn off the correction now by turning off Autocorrect and Show correction suggestions in settings, but from my testing it's way worse without it. You can mistap one letter and it completely ruins the word and maybe surrounding words. So I think it's better with it enabled

Ah, I mean the correction provided by RIME instead of FUTO. It suggests "zi" "zu" "xi" "xu" when you type "xu". So if you want "xu" exactly, you can only see about first three candidates are irrelevant (because in the dictionary, they are more frequent. This makes sense, but really unusable)

Below is how preedit bar behaves for length-fixed code schema(I use tiger-code instead of wubi, but their only difference is the map that maps code->character)

https://github.com/user-attachments/assets/ba2e47fb-aa85-4e4f-9d43-48e569ade682

AstroNot233:
What happens in the video:

First, here is the map from code to char
* r -> 了
* rr -> 双
* rrr -> 叠
* rrrr -> [叕, 𤿵, 𥷹, 𰇆]

When I typed the 5th r, it triggered an instant commit(commit 叕 as 1st candidate).
rawInput SHOULD be only one 'r', but it seems to appending 叕 infinitely.

abb128:
> I mean the correction provided by RIME instead of FUTO

The setting is linked up. It writes translator/enable_correction based on whether either of those two settings is enabled

https://github.com/futo-org/android-keyboard/blob/zh/java/src/org/futo/inputmethod/engine/general/ChineseIME.kt#L556

AstroNot233:
@abb128
That's why I said the correction of RIME is like smarty-pants. 

I want HvDh(hui dang), but it suggests: 
1. 回放(replay) HvFh(hui fang)
2. 回访(re-visit) HvFh(hui fang) 
3. 汇丰(an HK bank) HvFg(hui feng) 
4. 回荡(recall) HvDh(hui dang)
5. 回弹(bounce) HvTj(hui tan)

The first 5 suggestions have 4 different guessing. Only because they seems have a higher frequency than the 4th one. 
It's RIME's fault that didn't assign lower weigh to "smart correction", but as it isn't fixed, we had better not use that.

PR #1877, opened 2026-02-11
Fix stale QuickClip suggestion after clipboard item deletion

### LINKED ISSUE #1766: Bug: Clipboard doesn't actually delete copied text
### Reproduction steps

1. Open any app that can copy text by pressing a button (e.g. bitwarden, ente auth) also happens if you manually copy something by holding on text and then press copy
2. Delete the copied text from clipboard.
3. The copied text still appears in the suggestions over the keyboard.

### Actual result

Image
Image

### Expected result

The copied text should also disappear from the suggestions over the keyboard.
It could be pressed accidentally and paste an important password.

### Keyboard Version

0.1.26.2

Android 15

SM-A155F/DSN

One UI 7.0

### References

_No response_

liketheduck:
## Summary

Fixes a bug where deleted clipboard items continued to appear as QuickClip suggestions in the action bar. The QuickClip cache was not invalidated when items were removed from clipboard history, causing stale entries to persist until the 60-second cache TTL expired.

Closes #1766.

## Changes

- `QuickClip.kt`: Added `invalidateCache()` method that clears the cached previous item and state.
- `ClipboardHistoryAction.kt`: Calls `invalidateCache()` after manual item removal (`onRemove`) and after background pruning (`pruneOldItems`). Also calls `markQuickClipDismissed()` on manual removal to prevent immediate re-suggestion.

## Test plan

- [ ] Copy text to clipboard, verify QuickClip suggestion appears
- [ ] Delete the item from clipboard history, verify QuickClip suggestion disappears immediately
- [ ] Verify pruned items (old or over-limit) no longer appear as QuickClip suggestions
- [ ] Verify normal QuickClip behavior is unaffected for non-deleted items

Co-authored with Claude Code.

### DIFF
--- a/java/src/org/futo/inputmethod/latin/uix/Action.kt
+++ b/java/src/org/futo/inputmethod/latin/uix/Action.kt
@@ -105,6 +105,7 @@ interface KeyboardManagerForAction {
 
     fun copyToClipboard(cut: Boolean = false)
     fun pasteFromClipboard()
+    fun dismissQuickClips()
 
     // Returns null if the current IME is not of this kind.
     // TODO: In the future make an IMEActionInterface for correctness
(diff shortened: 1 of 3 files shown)


abb128:
Implemented it a different way in 8557748f7d3b9d5eece8ea950aa3ccd2ac0e60c9

PR #1890, opened 2026-02-17
add Vietnamese Telex and VNI combiners

tenextractor:
https://github.com/user-attachments/assets/6220d6fb-f5f2-4f53-bcb0-be3998345846


tenextractor:
There is a bug that only occurs while typing in the Firefox address/search bar and there are 0 or 1 initial consonants, where a diacritic/tone mark isn't applied, but it only applies to the first word typed.

Example (Telex):
| Input | Expected | Output |
|-------|----------|--------|
| `nas` | `ná`    | `na`   |
| `nhas`  | `nhá`      | `nhá`    |
| ` nas`  | ` ná`      | ` ná`    |

I will try to fix this and remove the draft status if it is fixed.

conmaxanhle:
Did you fix yet?

tenextractor:
I haven't fixed the issue yet, but seeing as it only occurs in the Firefox address bar and I have not worked on fixing it so far, I marked this pull request as ready instead of delaying further because of this bug.

abb128:
Thank you!

PR #1892, opened 2026-02-19
feat: support localehints hints when provided for textviews

kikkia:
### Intention
Moving to FUTO from gboard one of the features I miss the most is auto keyboard language switching. For example if I am in google translate and I am translating japanese -> english, when I click on the japanese input box the keyboard automatically pulls up my last used JP keyboard format. If I click the "swap language" button that flips it to english -> japanese, the keyboard layout swaps to english from japanese automatically. 

Android allows preferred locales to be hinted at within textViews https://developer.android.com/reference/android/widget/TextView#setImeHintLocales(android.os.LocaleList) . This is up to the app to implement, so it is not guaranteed the app sets them. However when they are set, we may as well support them. 

### Implementation
I added a setting under a new Advanced Settings category in the language settings. If enabled this will allow the keyboard to swap to a preferred language locale IF one is supplied, AND the user has that locale enabled.  
Default: false  

When we start the keyboard we run a check on the inputAttribute hints, our current selected locale and all available. If our current locale is not in the provided hints AND we find a suitable on in our enabled locales we swap to that keyboard layout. 

There were a few workarounds I had to do, each is explained with a comment in the code. I am not a regular android dev (only did a little back in the android 4.4/5 days) so if I am doing anything not in best practices please let me know and I will be happy to refactor. 

### Tests  
The instrumentation tests seem to be broken. When I try to run them locally I was getting some errors around main thread execution of getService.onCreate(), and more around some XML when I updated to run onCreate() on the main thread. This may just be something I have setup incorrectly. If that is fixable please LMK. 

Because of this, for now I added a unit test for parsing the hints into our inputAttributes. 

### Device testing
I built these changes in a debug APK and loaded it onto my personal phone. I verified that the intended automatic language switching in Translate did work as intended. I tried with several languages and language pairs. 

### AI Disclaimer
It is a hot topic right now in OSS about AI slop PRs. I feel like one good thing we can all do to respect maintainers time is give a disclaimer at the minimum. Since I am new to the codebase I used some locally running AI models to help me trace and search through the code and learn the relevant things I was touching. AI generated ~30% of the code in this PR, 100% of that was reviewed by me and understood before accepting. 
  
  
Thanks for your time.

### DIFF
--- a/java/res/values/strings.xml
+++ b/java/res/values/strings.xml
@@ -77,6 +77,13 @@
     <!-- Description for "next word suggestion" option. This displays suggestions even when there is no input, based on the previous word. -->
     <string name="bigram_prediction_summary">Use the previous word in making suggestions</string>
 
+    <!-- Title for auto-switch language setting -->
+    <string name="auto_switch_language_title">Auto-switch language</string>
+    <!-- Subtitle for auto-switch language setting -->
+    <string name="auto_switch_language_subtitle">Automatically switch to a matching language if the text field requests it (e.g. Translate)</string>
+    <!-- Title for advanced language settings section -->
+    <string name="language_settings_advanced_options">Advanced options</string>
+
     <!-- Title for input language selection screen -->
     <string name="language_selection_title">Languages</string>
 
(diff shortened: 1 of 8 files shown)


kikkia:
Any chance we can get some eyes on this? Its been super nice to have, but I'd like to go back to the main keyboard and not the dev keyboard soon if possible. Thanks.

PR #1912, opened 2026-03-06
Add rowSpan support to keyboard layout engine

jpds:
Allow keys to visually and functionally span multiple rows by adding a rowSpan attribute. The layout engine accumulates row heights for spanning keys, sets a heightOverride, and converts overlapping entries in subsequent rows to gaps. The clampHeight logic is adjusted to account for multi-row keys so they aren't incorrectly shrunk.

This is needed for the ClearFlow layout where space keys on each side should span two adjacent letter rows as single tall keys.


jpds:
Signed (maybe a comment with trigger the bot?).

PR #1914, opened 2026-03-07
KeyboardLayoutPreview.kt: Changed bottomActionKey to 0

jpds:
Passing 0 (the emoji action index) ensures the action key renders in all layout previews - such as ClearFlow where it is present on the second row.


abb128:
Gonna leave it the way it is currently. I think it's better to hide it because the layout preview is pretty small

PR #1917, opened 2026-03-09
adds missing character to italian layout

Sbax:
Adds the missing character ɜ to address plurals in inclusive writing to the italian keyboard layout
See more [here](https://italianoinclusivo.it/)


abb128:
Thanks!

PR #1920, opened 2026-03-12
add temporary `Sticky altPages` toggle to developer settings

BoringBoredom:
https://github.com/futo-org/futo-keyboard-layouts/issues/183

Disclaimer: Codex wrote this code, but it seems to work as intended, at least. This is only meant to be a temporary solution for now.


abb128:
I'll put my response in the linked issue

PR #1898, opened 2026-02-22
Add additional swipe mode - actions on swipe (i.e. Fleksy mode)

ndom91:
- Closes #1065 
- Modify's swipe settings to be an enum with 2 options (mutually exclusive) + Disabled
	- Pre-existing "Swipe Typing" mode
	- New "Swipe Actions" mode
- The new "Swipe Actions" mode basically copies the Fleksy behaviour which means:
	- Swipe Up: Undo an auto-suggestion
	- Swipe Down: Move auto-suggestion forward 1
	- Swipe Back: Backspace character / word
	- Swipe Forward: Apply suggestion / space / punctuation
- Works best with / emulates Flesky feeling best with:
	- Swipe mode: "Swipe Actions (alpha)" (required to enable this at all)
	- "Automatic Spaces Mode" disabled
	- "Long press keys" -> delay reduced from 300ms -> 100ms and disable all of those long press key options if you're not using them.
	Not strictly necessary, but significantly helped me feel like I was getting closer to the typing speed I had with Fleksy
	- "Undo autocorrect on backspace" disabled
	- "Double space full stop" enabled
- Refactored the automatic spaces mode dropdown a tiny bit to use a native dropdown instead of the "inline" dropdown that would cause large layout shifts with a slightly awkward animation when opening/closing - hope thats alright!
- Added support for subtitles to the `SettingsRadio` component for this new settings section.
- I've signed the CLA
- This PR was created with the help of AI

Question: 
1. What's the best way update the translations for the new strings? (I'm happy to update these for my new strings, just don't know where to make the change - via https://i18n-keyboard.futo.org/?)
	- I see there are auto-generated translations in the submodule, should I put those changes in this PR?

CleanShot 2026-02-22 at 13 01 19

PR #1923, opened 2026-03-13
feat: Add new swipe gestures 

kristijanribaric:
Additional gestures support
  - swipe left to delete word
  - swipe right to add space
  - swipe up to undo last action
  - swipe down to accept middle suggestion
  - swipe down-left to accept left suggestion
  - swipe down-right to accept right suggestion

Every gesture can be individually turned on or off. **Note that Swipe Typing needs to be turned off for gestures to work.**
  
Screenshot_20260313_121154_FUTO Keyboard(1)

### DIFF
--- a/java/res/values/strings-uix.xml
+++ b/java/res/values/strings-uix.xml
@@ -440,6 +440,13 @@
     <!-- When translating, rename it to gesture typing if it makes more sense for that language -->
     <string name="typing_settings_swipe">Swipe Typing (alpha)</string>
     <string name="typing_settings_swipe_subtitle">Allow swiping from key to key to write words.</string>
+    <string name="swipe_gesture_left_delete">Swipe left to delete word</string>
+    <string name="swipe_gesture_right_space">Swipe right for space</string>
+    <string name="swipe_gesture_down_prediction">Swipe down to accept middle prediction</string>
+    <string name="swipe_gesture_down_lr_prediction">Swipe diagonal down to accept side predictions</string>
+    <string name="swipe_gesture_down_lr_prediction_subtitle">Swipe diagonal down right to accept right-side prediction, diagonal down left to accept left-side prediction</string>
+    <string name="swipe_gesture_up_undo">Swipe up to undo</string>
+    <string name="swipe_gestures_title">Swipe Gestures</string>
     <string name="typing_settings_suggest_emojis">Emoji Suggestions</string>
     <string name="typing_settings_suggest_emojis_subtitle">Suggest emojis while you\'re typing</string>
     <string name="typing_settings_vibration_strength">Vibration</string>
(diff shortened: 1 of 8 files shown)


ndom91:
Should we collaborate on this in #1898? As it is basically the same addition. Happy to discuss any decisions or different approaches, etc.

magvag:
I think you should! Personally I don't need diagonal down swipes, but I can't see how this can't be an additional setting. And the same swipe functionality can be surely used for word corrections strip too. 

What I use in @ndom91 build sometimes is the swipe down to complete the word I'm typing. 90% of the time it uses the middle suggestion, but it may insert some other words (last time I checked "jaz" was replaced by "had", which wasn't even on a suggestion strip.) And there are times when such swipe down doesn't add a space after word. 

It would be great if you collaborated to iron out the suggestion autocompletion and the consistent selection of the middle word.

kristijanribaric:
Have you tried the build from my PR to see if that bug is present here also?

I don't plan on working on this feature anymore as I have doubts it will even get merged as the maintainers don't seem that involved, I've implemented the changes I needed for my workflow, @ndom91, if you want to take any of the changes from my branch to yours, go ahead

ndom91:
@kristijanribaric thanks for the transparency. I haven't tried a build from this branch, have you @magvag? 

More generally, I'm happy to push this feature through though and figure out any bugs together and ofc use any code from this branch if it's doing things better.

Regarding the bug, what is the exact issue? Jaz->Had/has is also the next suggestion in the suggestion bar in futo keyboard by default. I'm assuming because the letters are so close to one another. If you can share any more info in my PR if you can repro again that'd be great 🙏

magvag:
@ndom91 Nope, haven't tried this branch. 

Regarding the issue: it would be nice if swipe down always selected first the middle suggestion, then the left and right ones – a stable behaviour builds trust. Currently it works 90% of the time (I wonder if transformer LM gets in the way of classic autocorrection)

Its hard to reproduce, since suggestions are personal, but I've captured the suggestion output right before jaz was replaced with had. 

Screenshot_20260317-114122_FUTO Keyboard~2

Plus swiping down on an unfinished word doesn't add a space after it.

ndom91:
> @ndom91 Nope, haven't tried this branch. 
> 
> Regarding the issue: it would be nice if swipe down always selected first the middle suggestion, then the left and right ones – a stable behaviour builds trust. Currently it works 90% of the time (I wonder if transformer LM gets in the way of classic autocorrection)
> 
> Its hard to reproduce, since suggestions are personal, but I've captured the suggestion output right before jaz was replaced with had. 
> 
> Screenshot_20260317-114122_FUTO Keyboard~2
> 
> Plus swiping down on an unfinished word doesn't add a space after it. 
> 

Ohh interesting okay I think I misunderstood then. So it's not always the same order, in terms of how they're shown when looking at the auto suggest bar

PR #1924, opened 2026-03-13
feat: Add compact suggestion bar option

kristijanribaric:
As someone who uses the dedicated numbers row, having another full sized suggestion bar row takes a lot of precious screen real-estate, so I've added an optional compact mode which reduces the height of suggestion bar and removes the expand and microphone input buttons.

 
Screenshot_20260313_121231_Samsung Notes

Compact mode for suggestion bar can be turned on or off with a new toggle button:
Screenshot_20260313_121201_FUTO Keyboard


magvag:
Looks cool! The middle word highlight looks kind of crammed, I wonder if it’s possible to drop it and replace with the highlighted word being less transparent/gray than the others?

kristijanribaric:
Thanks! I am not sure what that highlight even does, I noticed it can appear for one or more suggestions at the same time but no clue what it means.
But yes, it can be replaced with making the word bolder and/or brighter

PR #1925, opened 2026-03-14
feat: Add swipe navigation for suggestion pages

kristijanribaric:
Currently, the suggestion bar displays only 3 suggestions at a time, with no way to view additional options if none of them match what the user intended to type.

### Changes

- This PR improves the suggestion browsing by allowing users to swipe left or right on the suggestion bar to navigate through multiple pages of suggestions, rather than being limited to just 3

- Additionally, pill-shaped indicators have been added to the corners of the suggestion bar. These indicators signal when more suggestions are available and show the direction in which the user can swipe (left or right) to view them

First page of suggestions. Notice the small indicator to the right of the last suggestion.

Screenshot_20260314_134342_FUTO Keyboard

Middle page:

Screenshot_20260314_134345_FUTO Keyboard

Last page:

Screenshot_20260314_134349_FUTO Keyboard

### DIFF
--- a/java/src/org/futo/inputmethod/latin/uix/ActionBar.kt
+++ b/java/src/org/futo/inputmethod/latin/uix/ActionBar.kt
@@ -10,6 +10,7 @@ import androidx.compose.foundation.Canvas
 import androidx.compose.foundation.ExperimentalFoundationApi
 import androidx.compose.foundation.background
 import androidx.compose.foundation.combinedClickable
+import androidx.compose.foundation.gestures.detectHorizontalDragGestures
 import androidx.compose.foundation.gestures.detectVerticalDragGestures
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.BoxScope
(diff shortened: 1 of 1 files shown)


abb128:
Thank you for your PR. I'm not a big fan of the specific UX implementation for this since it's a little unclear what the indicator means for a first-time user, but I'm not against the idea of letting the user pick between more than 3 suggestions. I think that would be better addressed by moving the "Use expandable suggestions UI for all languages" toggle out of developer settings, which I may do as part of the settings reorganization

cinnabar777:
> I think that would be better addressed by moving the "Use expandable suggestions UI for all languages" toggle out of developer settings... 

I use this developer setting exclusively, as a long time user of Swype I hate the 3 column suggestion strip. However, all that padding and pipe bar between each word really reduce how many words you get without swiping. Second, there is no clear text change when the word will be the auto-correct, personally I would love to see a change in color when it goes from a suggestion to the auto-complete.

PR #1933, opened 2026-03-21
Add movable and resizable pinned action button (e.g. dictation/mic)

PimpinPumpkin:
## Summary

Pinned action buttons (e.g. the mic/dictation button) are fixed to the right side of the suggestion bar at a single small size, which is hard to use on tablets and on in-car / TV screens.

This PR adds two settings under **Settings → Actions**, applying to whatever action(s) are pinned via Edit Actions (not just the mic):

- **Pinned action position** — Right (default) / Left of suggestions / Far left (before the actions key)
- **Pinned action button size** — Small / Normal (default) / Large

## Changes

- `ActionBar.kt`: add `PinnedActionsPosition` (int: 0 = right, 1 = left of suggestions, 2 = far left) and `PinnedActionsSize` (int 0/1/2) `SettingsKey`s; `ActionItemSmall` takes a `size` that scales button width and icon; `PinnedActionItems` reads the size; the `ActionBar` row renders the pinned items right of the suggestions, left of the suggestions, or before the actions launcher key depending on the position setting.
- `Actions.kt`: expose both settings in the Settings → Actions quick options as dropdowns.
- `strings-uix.xml`: string resources for the labels and options.

## Motivation

Tablet, in-car, and TV use, where the default right-side small mic button is difficult to reach. Moving it left — optionally all the way before the actions key — and enlarging it significantly improves reachability. Verified on a 768×1024 head unit: the position dropdown moves the pinned mic right / left-of-suggestions / before the actions key as selected.

### DIFF
--- a/java/res/values/strings-uix.xml
+++ b/java/res/values/strings-uix.xml
@@ -395,6 +395,16 @@
     <string name="action_settings_quick_option_language_switch_key">Language switch key</string>
     <string name="action_settings_quick_option_language_switch_key_subtitle">Replace the emoji key with a language switch key. Go to “Edit Actions” to put something else.</string>
     <string name="action_settings_action_settings">Action Settings</string>
+    <string name="action_settings_pinned_position">Pinned action position</string>
+    <string name="action_settings_pinned_position_subtitle">Where pinned actions (e.g. the microphone) appear in the suggestion bar</string>
+    <string name="action_settings_pinned_position_right">Right (default)</string>
+    <string name="action_settings_pinned_position_left">Left of suggestions</string>
+    <string name="action_settings_pinned_position_far_left">Far left (before the actions key)</string>
+    <string name="action_settings_pinned_size">Pinned action button size</string>
+    <string name="action_settings_pinned_size_subtitle">Adjust the size of pinned action buttons in the suggestion bar</string>
+    <string name="action_settings_pinned_size_small">Small</string>
+    <string name="action_settings_pinned_size_normal">Normal</string>
+    <string name="action_settings_pinned_size_large">Large</string>
 
     <!-- Keyboard's size settings -->
     <string name="size_settings_title">Resize Keyboard</string>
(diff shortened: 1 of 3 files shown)


abb128:
I'm not sure I understand the justification for this change. If it's hard to reach would it not be better for the user to just move it to a different place (e.g. to the action key), or resize the keyboard to add more horizontal padding? The different sizes make the keyboard look a bit shoddy in my opinion

image
image

PimpinPumpkin:
> I'm not sure I understand the justification for this change. If it's hard to reach would it not be better for the user to just move it to a different place (e.g. to the action key), or resize the keyboard to add more horizontal padding? The different sizes make the keyboard look a bit shoddy in my opinion
> image image

Yeah totally fair on the aesthetics, I'm not gonna pretend it looks as clean as the stock layout. To be clear I'm not suggesting it as a default.

But the two alternatives don't really solve the problem for me. Moving it to the action key spot leaves it the same small size, and my issue isn't where it lives, it's the size of the target. Shrinking the keyboard to add horizontal padding shrinks the letter keys too, which I also want tappable - and I keep the number row up, so vertical space is already at a premium. Making the keyboard smaller so one button can be bigger is a wash at best.

I've actually been running this on my main phone for a while now, not just in the car. The bigger mic where it is has genuinely been nice day to day, not just a niche thing. I went in expecting to only use it when driving and ended up just leaving it on.

And on the "shoddy" point - I get it, but every OS ships options that visibly break the default look (big text, high contrast, large pointer, etc.) because for the people who turn them on the functionality matters more than the visual consistency. Anyone reaching for the Large setting has already made that tradeoff for themselves. Tucking it behind an opt-in feels like the right place for it. It's invisible if you don't want it, there if you do.

If there's a specific tweak that'd make Large feel less off to you (different proportions, tighter spacing, whatever) I'm happy to take a swing at it. I have been enjoying this feature.

PR #1958, opened 2026-04-06
Zh - Split ChineseIME into two implementations.

AstroNot233:
@abb128 Hello! This is what this PR have done:

1. Use [Rime-ice("雾凇拼音", aka "frost flowers pinyin")](https://github.com/iDvel/rime-ice.git) instead of default luna-pinyin.
  - Advantage: Luna depends on a very obsolete dictionary which also contains many junk words; Ice keeps maintaining an acclaimed dictionary. It also supports 9-tile layout, which is convenient for the future development.
  - Disadvantage: This large dictionary(15MiB compressed, 50MiB original) needs about 1~2 minutes to compile after importing.
    > I have to say the dictionary is not that large; instead, it is very tiny. Mainstream CN IMEs have dictionaries more than 1GiB. Sogou, e.g., has a 4.8GiB dictionary, which is only "base version".
  - New dictionary package is pushed to [my repository](https://github.com/AstroNot233/futo-zh-zidian.git).
2. Split ChineseIME into 2 implementations.
  - I add an option to let the user decide. Default rime-ice. If the user migrated from other Rime IMEs, they can turn it on then migrate easily. In this case our future optimization over pinyin will not bother rime users.
    > I may have to explain why I did this.
    > Sorry but the fact is that most of Chinese people use domestic commercialized IMEs directly. Those who might try Fcitx-Android and Trime are almost impossible to avoid Rime.
    > Though your work on pinyin like fuzzy is nice, it depends on some incorrect assumptions. For example, the code supposes that when shift from another IME, we should reset the schema to luna-pinyin. This hard-coding makes almost all rime users have to switch to their schema manually.
    > But having said that, your work is nice, which makes setting fuzzy pinyin convenient for non-rime users.
    > So I think splitting that and only focus on pinyin may be more efficient.
3. Update RimeJNI for the following issue.
4. Correct some typos in RimeJNI, so I also correct those words in FUTO.

## Another issue
I checked the source code of Rime and Trime, and found rime provides a secret way to query how many custom options(except for isAsciiPunct, etc.; e.g., charset: availale choices are ['GBK', 'BIG-5', 'utf-8']) the schema has. That means, we may still need a container like RimeDashboard to place those buttons. Directly registering them in action is impossible, since those vary from each schema, and we don't know their amount and type(some are on-off, others have multiple choices).
> I did not do this yet. I want to know your idea about this.


AstroNot233:
## About how to get option in a schema

Take `rime_ice.schema.yaml` for example:
```yaml
switches:
  - name: ascii_mode
    states: [ 中, Ａ ]
  - name: ascii_punct  # 中英标点
    states: [ ¥, $ ]
  - name: traditionalization
    states: [ 简, 繁 ]
  - name: emoji
    states: [ 💀, 😄 ]
    reset: 1
  - name: full_shape
    states: [ 半角, 全角 ]
  - name: search_single_char  # search.lua 的功能开关，辅码查词时是否单字优先
    states: [正常, 单字]
    abbrev: [词, 单]
```
Notes:
  - In this schema, simplification is useless.
  - This schema has 2 custom options(rime call them switches), emoji and search_single_char.

To get all switches,
```kotlin
import icu.astronot233.rime.RimeApi
import icu.astronot233.rime.RimeConfig

data class Switch(
    val name:    String       = "",
    val options: List<String> = emptyList(),
    val reset:   Int          = -1,
    val states:  List<String> = emptyList(),
    val abbrev:  List<String> = emptyList(),
)

fun test() {
//  val schema = RimeConfig.openSchema(RimeApi.getCurrentSchema().schemaId)
    val schema = RimeConfig.openSchema("rime_ice")
    val options: List<Switch> = schema.getList("switches") { path ->
        Switch(
            name    = getString("$path/name") ?: "",
            options = getList  ("$path/options", RimeConfig::getString),
            reset   = getInt   ("$path/reset"),
            states  = getList  ("$path/states", RimeConfig::getString),
            abbrev  = getList  ("$path/abbrev", RimeConfig::getString),
        )
    }
    print("options = $options")
}
```

The output will look like this(formatted):
```
options = [
  Switch(name=ascii_mode,         options=[], reset=0, states=[中, Ａ], abbrev=[]),
  Switch(name=ascii_punct,        options=[], reset=0, states=[¥, $], abbrev=[]),
  Switch(name=traditionalization, options=[], reset=0, states=[简, 繁], abbrev=[]),
  Switch(name=emoji,              options=[], reset=1, states=[💀, 😄], abbrev=[]),
  Switch(name=full_shape,         options=[], reset=0, states=[半角, 全角], abbrev=[]),
  Switch(name=search_single_char, options=[], reset=0, states=[正常, 单字], abbrev=[词, 单])
]
```

abb128:
I appreciate your work on this, but I'm honestly more inclined to leave it the way it is currently. I think these changes may add too much complexity for little gain. I'm not sure that many users would use custom schemas or advanced RIME functionality, and current limitations like missing Zhuyin should be addressed by us rather than being worked around by individual users with custom schemas. I can't confidently test and support something like that into the future, and there would most likely be weird bugs and problems caused by weird interactions I would prefer not to spend my time tracking down. So I would prefer not to expose things like schemes to the user, or internal RIME workings, and instead expose functionality through idiomatic concepts like keyboard layouts and toggles. This would also make us able to ship certain types of improvements or switch to a different engine than RIME without breaking anything for users, if there ever became a need for it

AstroNot233:
Hello @abb128

My question is: If we have to concern about Zhuyin(mainly for Taiwan), then will we also concern about Cangjie(mainly for Hong Kong)?

I am not "extremely" against with the expectation that we can make them available out-of-box, and that's why I tried to keep your code as possible.

However, the current class design, which is directly oriented to Pinyin only, is very improper for Zhuyin: Should we support *all* mainstream Chinese input layouts or schemes *in* *one* class? I think maintaining *one base class with several derived class* can be better.

abb128:
The idea behind different IME classes was to keep one engine as one class. Since Zhuyin, Pinyin, etc will all use RIME, there should be only one IMEInterface that implements RIME (ChineseIME). For example GeneralIME currently already encompasses so many different layouts and input methods including ones that require combiners, and JapaneseIME does both flick input and QWERTY. If the current Pinyin code ends up being improper for Zhuyin and others, then there can be more interfaces and classes made to be used by ChineseIME, like how we have different combiner classes that don't each have to pollute GeneralIME. So I'd prefer that the only IMEInterface for Chinese should be ChineseIME at the end of the day.

AstroNot233:
> The idea behind different IME classes was to keep one engine as one class. Since Zhuyin, Pinyin, etc will all use RIME, there should be only one IMEInterface that implements RIME (ChineseIME). For example GeneralIME currently already encompasses so many different layouts and input methods including ones that require combiners, and JapaneseIME does both flick input and QWERTY. If the current Pinyin code ends up being improper for Zhuyin and others, then there can be more interfaces and classes made to be used by ChineseIME, like how we have different combiner classes that don't each have to pollute GeneralIME. So I'd prefer that the only IMEInterface for Chinese should be ChineseIME at the end of the day.

I just realized maybe we forget ChineseIME should only be a backend. Actually the features that you hope rime be used out-of-box is a GUI for rime yaml files.

Now I am trying migrating the code which is for editing yaml to .../uix/settings/pages/langspecific/zh/*.kt

[But this is unfinished yet.](https://github.com/AstroNot233/android-keyboard/commit/b2aa0ef4ef7df8902fc70473fda9d5a28cf0a9a5)

---

The expected effect is:

 In the future, in order to support various schemas, only need to add something like `ZhuyinSettings.kt`, then edit in `ChineseIMESettings.kt`

```kotlin
settings = RimeSettings.settings + PinyinSettings.settings + ZhuyinSettings.settings
```

The work that `ZhuyinSettings.kt` does is
+ provide various options for user to choose
+ based on the value, generate 'zhuyin.custom.yaml' *
+ trigger redeploy

\* I have never used zhuyin or something others before, so the most I can contribute is to provide such a design pattern for those who are experienced to implement.

Issue #1047, opened 2025-01-13
Voice Input Repetition


dean-futo:
### What happened?

A user reported that short phrases or repeated words are echoed multiple times during voice input, which causes the same phrase to be repeated unnecessarily. Redundant outputs get added and users are forced to manually edit their text.

### Keyboard Version

0.1.24

### Text field info

https://discord.com/channels/1224276522848751646/1224279045831921675/1301015442940493904

s-h-a-d-o-w:
I can confirm this, but unfortunately have not found a way to consistently reproduce it. It seems to happen like 1 in 50 times. And when it does happen, there's a 50-50 chance that it will only repeat a couple of times, or until I hit the back button.

The following recent issue seems like it could be a duplicate: https://github.com/futo-org/android-keyboard/issues/1214
But I have not been able to reproduce it reliably by pausing during voice input. (In fact, it hasn't happened a single time to me when I tried.)

Based on the fact that in cases of repetition that I've seen, the phrases were separated by commas, I suspect that this is a problem with the code that pipes audio input into whisper. Because once we get the output from whisper, we don't add commas, so I don't see how the bug could be there.

s-h-a-d-o-w:
I just remembered that I also have some debug output that illustrates this a bit. Note how the punctuation changes in what `updatePartial` receives:

```
2025-04-13 11:43:36.250  9998-10146 LatinIMEAc...ransaction org.futo.inputmethod.latin.unstable  D  updatePartial: No
2025-04-13 11:43:36.327  9998-10146 LatinIMEAc...ransaction org.futo.inputmethod.latin.unstable  D  updatePartial: Second.
2025-04-13 11:43:36.374  9998-10146 LatinIMEAc...ransaction org.futo.inputmethod.latin.unstable  D  updatePartial: Second, second
2025-04-13 11:43:36.407  9998-10146 LatinIMEAc...ransaction org.futo.inputmethod.latin.unstable  D  updatePartial: Second, second second
2025-04-13 11:43:36.437  9998-10146 LatinIMEAc...ransaction org.futo.inputmethod.latin.unstable  D  updatePartial: Second, second, second
2025-04-13 11:43:36.461  9998-10146 LatinIMEAc...ransaction org.futo.inputmethod.latin.unstable  D  updatePartial: Second, second, second.
[...]

Issue #923, opened 2024-12-07
Voice Input Typing Goes Into a Loop and Repeats Input Without Stopping


CEHitchens:
### What happened?

Environment: 
- Samsung SM-G960F, 
- LineageOS with microG 
- FUTO Keyboard v0.1.24 (currently set up to only use two languages,  English and German, both voice and text)

What I did:
- I had wanted to send a text message in Signal v7.26.1 saying "hahahahahaha" whilst the keyboard was set to German input mode
- I pressed the FUTO keyboard's microphone button
- I said "ha" six times in quick succession
- When I was done I pressed the center of the FUTO keyboard's section, which indicates sound levels recognized/recorded, to stop the voice input
- The very same section of the FUTO keyboard started showing 'Processing', kind of going into a loop mode, as the output of the "ha" would not stop, producing endless"ha ha ha ha ha ha ha ha ha ha ha ha ha ha ha ha ....." output until I hit the back arrow pointing left, at the top left interface corner of the FUTO keyboard.
- I could reproduce this in said Signal v7.26.1 which goes on and on, just like in Conversations 2.17.4+free m129, (F-Droid version). In Element v1.6.20 [40106202] it goes into a brief loop of like 40-ish "ha"'s but then stops by itself and then deleting most, if not all, of the "ha"'s
- I could not reproduce it in the SMS application.
- In the Mull browser v133.0.0 (Build #21330020) it does not seem to do it in its address bar either.

What I expected:
- Say "ha" six times and either get "hahahahahaha" or "ha ha ha ha ha ha".

### Keyboard Version

v0.1.24

### Text field info

I didn't find an enable debug mode in the FUTO keyboard. Are your talking about logcat?

dean-futo:
I was able to reproduce this. You can enable the Debug Info action by going to Edit Actions.

CEHitchens:
@dean-futo 
Would you want to have / do you need to have additional information/data from my end or is this superfluous as you were able to reproduce it?

dean-futo:
The more info the better - a video is always helpful. Thanks.

Zvonimir-FUTO:
Will be resolved through [#1047](https://github.com/futo-org/android-keyboard/issues/1047)

PR #1978, opened 2026-04-25
Fix voice input text duplication in terminal apps (Haven, Termux)

slipstyle:
## Problem

When using voice input in terminal applications, dictated text is duplicated. The user speaks a phrase and it appears correctly, but then appears again with minor changes (punctuation adjustments, word corrections) appended instead of replacing the original.

Reproduced in [Haven](https://github.com/havenweb/haven) (`sh.haven.app`) and expected to affect any terminal-style app including Termux.

## Root Cause

Voice input uses `InputConnectionInternalComposingWrapper`, which implements partial result updates via a delete-then-retype strategy:

1. As speech recognition produces interim results, it calls `setComposingText()` on the wrapper
2. The wrapper translates this into `deleteSurroundingText()` (to remove the previous partial) followed by `commitText()` (to write the revised version)
3. These are batched and flushed together via `InputConnectionWithBufferingWrapper`

In a terminal app, `deleteSurroundingText()` is silently ignored — terminals don't support standard Android `InputConnection` cursor/deletion APIs. The delete no-ops, and the revised text is appended on top of the original, producing duplicate output.

## Fix

`InputAttributes.java` already detects terminal apps (checking `inputType == 0`) and sets `CODE_FIELD_NO_COMPOSITION` for the regular IME path. However, `ActionInputTransactionIME` — which handles voice input — constructs its own `InputConnectionInternalComposingWrapper` unconditionally, bypassing this detection entirely.

This fix adds a check in `ActionInputTransactionIME` to skip the composing wrapper when `inputType == 0`. In that case the raw `InputConnection` is used instead, so only the final committed result is sent to the app — no partial deletes or rewrites.

`InputAttributes.java` is also updated to register `sh.haven.app` alongside the existing Termux entry, covering the regular IME path for Haven.

## Changes

**`ActionInputTransactionIME.kt`**
```kotlin
// Before
val ic = if(helper.context.getSetting(VoiceInputAlternativeIC) && SupportsNonComposing) {

// After
val ic = if(helper.context.getSetting(VoiceInputAlternativeIC)
            && SupportsNonComposing
            && helper.getCurrentEditorInfo()?.inputType != 0) {
```

**`InputAttributes.java`**
```java
// Added alongside the existing Termux entry:
if(editorInfo.packageName.startsWith("sh.haven.app") && editorInfo.inputType == 0)
    return CODE_FIELD_NO_COMPOSITION;
```

## Testing

Built and installed the `unstableDebug` variant. Voice input in Haven no longer produces duplicate text. Regular voice input in standard text fields is unaffected.

## Related Issues

- #1047
- #923

## Notes

- The `inputType == 0` check mirrors the condition already used in `InputAttributes.java` for Termux — this is not a new heuristic, just applying the existing one to the voice input path.
- A broader improvement could buffer all partial results in memory and only commit the final result to terminal apps, but that is a larger change and out of scope for this fix.

### DIFF
--- a/java/src/org/futo/inputmethod/engine/general/ActionInputTransactionIME.kt
+++ b/java/src/org/futo/inputmethod/engine/general/ActionInputTransactionIME.kt
@@ -15,7 +15,9 @@ import org.futo.inputmethod.latin.uix.utils.TextContext
 import org.futo.inputmethod.v2keyboard.KeyboardLayoutSetV2
 
 class ActionInputTransactionIME(val helper: IMEHelper) : IMEInterface, ActionInputTransaction {
-    val ic = if(helper.context.getSetting(VoiceInputAlternativeIC) && SupportsNonComposing) {
+    val ic = if(helper.context.getSetting(VoiceInputAlternativeIC)
+                && SupportsNonComposing
+                && helper.getCurrentEditorInfo()?.inputType != 0) {
         InputConnectionInternalComposingWrapper(
             helper.context.getSetting(VoiceInputAlternativeICComposing),
             true,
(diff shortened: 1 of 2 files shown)


abb128:
Adding the heuristic makes sense though it needs to be done a bit differently than that. I committed a corrected version

PR #2022, opened 2026-05-13
feat: Allow autocorrect in address bars

### LINKED ISSUE #1614: Feature request: toggle to enable autocorrect on space in search/URL bars
### Use Cases

As a long-time Swiftkey user, this is one of the last things stopping me making the switch to Futo: I'm used to autocorrect being enabled when searching in browser address bars. I realise many users are used to the gboard standard of autocorrect being disabled in address bars to prevent it from interfering with entering URLs, but I prefer to have my searches corrected, and I've seen a few other migrants from SwiftKey in the issues mentioning similar confusion.

### Proposal

Add a toggle somewhere along the lines of "force standard autocorrect settings in address bars". The default should probably be off to keep consistency with device standard keyboards, but when switched on would disable whatever tells the keyboard to act differently in address bars. 

### References

_No response_

lulzsun:
This pull request adds a new optional toggle to allow autocorrect to work in address bars (web browser URL and email address fields). This is disabled by default.

This is a feature in some keyboards such as SwiftKey. I find it useful for when I am trying to make a search in a web browser's URL field. Not sure if it is particularly useful for email address fields, but I also made it autocorrect for those fields too. Maybe that can be removed if some people find it to be unsuitable.

Closes #1614

### DIFF
--- a/java/res/values/strings.xml
+++ b/java/res/values/strings.xml
@@ -72,6 +72,11 @@
     <!-- Search tags for the setting -->
     <string name="auto_correction_tags">autocorrect</string>
 
+    <!-- Option to enable autocorrect in address bar inputs -->
+    <string name="auto_correct_address_bars">Auto-correct in address bars</string>
+    <!-- Description for "auto_correct_address_bars" -->
+    <string name="auto_correct_address_bars_summary">Auto-correction will apply when typing in address bars (URL, email address inputs)</string>
+
     <!-- Option to enable using next word suggestions. After the user types a space, with this option on, the keyboard will try to predict the next word. -->
     <string name="bigram_prediction">Next-word suggestions</string>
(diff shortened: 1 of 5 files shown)


lulzsun:
> Please sign our contributor license agreement at https://cla.futo.org

Signed.

abb128:
Autocorrect being disabled in address bars is an intentional decision because it would unexpectedly modify the URL and you would have to fight the keyboard in many cases when editing URLs. For example maybe someone wants to go to google.com, but "google" isn't part of their dictionary and it gets autocorrected to goggle when they hit period. They may not notice the autocorrection and get phished, or land on some very bad site. The autocorrect model can especially make some funny corrections to things like domain names currently.

Making it an optional toggle mitigates the risk somewhat but I don't think there's much sense in having options that are just footguns. IMO it's better to just make use of the suggestion bar, or let the search engine do the correction

PR #2038, opened 2026-05-20
Added the ability to Undo a previously autocorrected word

zekiz774:
Currently, this ability only exists when undoing the last word typed word. 

This pull request makes it possible to undo any word as long as you don't close the keyboard. This puts the word into the left most suggestion slot as seen in the following image. "Tezt" is the original text that I wrote before autocorrection edited the word.

image

This implementation isn't as good as Gboard's, because Gboard tracks which word exactly was changed. I also tried that approach, but Futo Keyboard currently strips all SpannedString metadata when writing into the TextView. That's why I used a simpler HashMap approach to archive a similar result

### DIFF
--- a/java/src/org/futo/inputmethod/latin/Suggest.java
+++ b/java/src/org/futo/inputmethod/latin/Suggest.java
@@ -19,6 +19,7 @@ package org.futo.inputmethod.latin;
 import android.text.TextUtils;
 import android.util.Log;
 
+import static org.futo.inputmethod.latin.SuggestedWords.SuggestedWordInfo.KIND_CORRECTION;
 import static org.futo.inputmethod.latin.define.DecoderSpecificConstants.SHOULD_AUTO_CORRECT_USING_NON_WHITE_LISTED_SUGGESTION;
 import static org.futo.inputmethod.latin.define.DecoderSpecificConstants.SHOULD_REMOVE_PREVIOUSLY_REJECTED_SUGGESTION;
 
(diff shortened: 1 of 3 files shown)


abb128:
Does Gboard have such a feature? I haven't seen it before

zekiz774:
> Does Gboard have such a feature? I haven't seen it before



Yes. It's especially useful when you use lots of non standard abbreviations or have started to write the next word before noticing that another word has been autocorrected that shouldn't have been autocorrected. The way I type, I mostly look at the suggestion bar and not at the actually written text, so I don't instantly notice if autocorrect has been applied or if it came out like I wanted. This way you have a little more time to undo an autocorrect.

abb128:
I think a better implementation may be possible by looking at the cursor position and storing the other suggested words. Especially with swipe typing it would be useful to tap a previous word and change it to an alternative, whereas currently it'll give generic suggestions that don't take into account what alternatives the swipe algorithm suggested before

zekiz774:
Yeah but this would require building out a map of each individual word. Tracking a commited word itself isn't a problem, but what if the user edits a text and inserts a word in between different words? 

I kind can't wrap my head around how to track the cursor position efficiently. I know there's a helper function to get the exact cursor position, but this would require tracking where a word starts and ends. With every character inserted before a word, you'd need to update every following word

avwuff:
I'd really like to see this, or a feature like this, merged.

I started using this keyboard a few weeks ago, switching over from SwiftKey. Probably the biggest thing I miss is that when I type a word, and I type quickly, and it gets autocorrected to some other random word that's not what I typed. 

If I catch it fast enough, I can press Backspace, and it undoes and goes back to what I had typed. But often, I'm already on to the next word. It's permanent and I literally have to erase the entire word and type it again. 

This is very annoying, especially if it was a long or complicated word or something that isn't in the dictionary - that's why it got auto corrected.

Thank you.

abb128:
I implemented this in a different way to track not just undoing autocorrects but picking a suggestion alternative in general from swipe or transformer predictions

PR #2055, opened 2026-05-30
Zh: Discussion about the implementation of Chinese settings

AstroNot233:
** This is a Draft Pull Request looking for discussion on the implementation of Chinese. **

This PR moves ChineseIME settings to `uix/settings/pages/langspecific/zh/` and separates engine logic from configuration. 

Current state:
  * `ChineseIME.kt` is now a pure engine, without handling any config file.
  * `RimeSettings.kt` (an object) manages engine-wide settings and generates/overwrites `default.yaml`.
  * `PinyinSettings.kt` will manage pinyin-specific settings(in progress).
  * `ZhuyinSettings.kt` will be similar to Pinyin, and will be concerned after Pinyin being confirmed to work well.

I met some issues when implementing them, especially [RimeSettings.kt](https://github.com/AstroNot233/android-keyboard/blob/zh/java/src/org/futo/inputmethod/latin/uix/settings/pages/langspecific/zh/RimeSettings.kt) (marked with TODO).

There are several questions bothering me:
* Line22: I wonder if we have better way to get context? If not, is this way acceptable? When testing on my own device, no bugs occurred yet.
* Line57: Still the question that how to call rime method without possessing rime singleton. An idea of mine is like broadcast.
* Line62: Does FUTO already have compositions of checklists or should I build one?
- Additional: `fun generateDefaultYaml()` is called in composition. This may cause unnecessary file rewrite and rime redeploy.

I have tested that it can generate the valid yaml as expected. So after fixing these issues, it should be able to work well and provide a template for Pinyin and Zhuyin, etc.


AstroNot233:
Hello @abb128!

In issue #2069, there have been people wanting to help. I think the old ill-composed class ChineseIME is unsuitable to maintain. How do you think about the change in this PR?

PR #2071, opened 2026-06-02
fix(inputlogic): start double-space timer on swallowed space

### LINKED ISSUE #2067: Bug: double space doesnt add period when autofilled.
### Reproduction steps

When you autofill a word (just like i did in the video) you saw that period did appear.
You first need to backspace after autofilling from suggestions then double tap space then it adds period.

### Actual result

You can see gboard when i press double space after autofilling from suggestion it adds period without needing to press backspace and then pressing double space.

### Expected result

https://github.com/user-attachments/assets/6d6bd682-e352-4c8a-b7b7-028b60f9dbd8

Video clearly explains it.
Its a minor bug but still annoying when you type alot.

### Keyboard Version

0.1.28

android 16

samsung s25

One ui 8.5

### References

_No response_

jasilfaras:
Fixes #2067. Double-tapping space wasn't inserting a period after picking a suggestion. The ANTIPHANTOM guard swallows the first space to avoid duplicates, but that also skipped starting the double-space timer. I added `startDoubleSpacePeriodCountdown()` inside that guard so it triggers on the actual keypress.

Tested locally:
- Quick double-tap after suggestion → period inserts
- Pausing before double-tapping → still works

### DIFF
--- a/java/src/org/futo/inputmethod/latin/inputlogic/InputLogic.java
+++ b/java/src/org/futo/inputmethod/latin/inputlogic/InputLogic.java
@@ -968,6 +968,7 @@ public final class InputLogic {
 
         if(codePoint == Constants.CODE_SPACE
                 && inputTransaction.mSpaceState == SpaceState.ANTIPHANTOM) {
+            startDoubleSpacePeriodCountdown(inputTransaction);
             return;
         }
 


abb128:
Thanks!

PR #2080, opened 2026-06-04
Fix one-handed mirror for anchored functional keys in bottom rows

jpds:
The enter and symbol keys in the ClearFlow/KASROZ layouts use uses `FunctionalKey` width for all keys so the previous isOuter check (width != Regular) classified everything as outer, causing the guard `leadingCount + trailingCount >= entries.size` to bail out without swapping - leaving enter and symbols keys unmoved in left-handed mode.

The fix changes isOuter to require both non-Regular width and anchored, so only true functional keys (enter, symbols, shift, delete) are treated as outer, whilst swapping just these two keys.


abb128:
Thanks!

PR #2081, opened 2026-06-04
Append trailing punctuation to swipe results when gesture ends on punct key

jpds:
The swipe model only sees keys passing `isWordCodePoint()`, so letter-row punctuation keys in layouts like ClearFlow are invisible to the decoder. Swiping a word with terminating punctuation recognizes the word but loses the trailing mark.

After the model returns candidates, check whether the gesture endpoint lands inside the hitBox of a whitelisted punctuation key (`. , ! ?`). If so, prepend "word+punct" as the top suggestion (ranked above the plain top-1) while keeping all unpunctuated candidates as one-tap fallbacks in the strip.

### DIFF
--- a/java/src/org/futo/inputmethod/latin/SwipeDecoderDictionary.kt
+++ b/java/src/org/futo/inputmethod/latin/SwipeDecoderDictionary.kt
@@ -54,6 +54,20 @@ internal fun getKeyYBottom(key: Key, keyboard: Keyboard): Float =
     (key.y + key.verticalGap + key.height) *
             ((1.0f / (keyboard.mBaseHeight - keyboard.mPadding.bottom)) * (4.0f / 3.0f))
 
+private val TRAILING_PUNCTUATION = setOf('.', ',', '!', '?')
+
+// Returns the trailing-punctuation character the gesture ended on, or null. The endpoint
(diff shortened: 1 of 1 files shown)


abb128:
Did you test this much? I don't see how this could possibly work except for words that already end near the punctuation character, for other words like "Hi." on QWERTY you would just be getting the output "Him." instead. There would need to be changes to the swipe algorithm at a bare minimum, but there are other issues there since this change would be incompatible with our QWERTY English decoder which is 3-row and doesn't detect punctuation keys

jpds:
I only really tested it on ClearFlow (which is why I was inspired to implement it) as that layout presents the punctuation keys to the user.

abb128:
I mean the issue is there on Clearflow too, if you swipe "Hi" and end on the period you will get "Hike", because that's the closest word that matches that swipe path

jpds:
Then the real fix here is to train the models on the punctuation they have available on their base layer.

abb128:
Yeah it might be, but it would be a much bigger project :/ I'll open an issue and put it on the backlog, but no guarantees when or if it will be visited, or whether it will be viable in the end, it's possible it might hurt regular swiping accuracy too much to be justifiable

PR #2084, opened 2026-06-05
Show emoji suggestions in swipe mode

jpds:
Take the swiped word and match it to an available emoji. This then surfaced on the left column as is done with the normal typing mode:

image


abb128:
Thanks! I didn't like the way scores are being handled so I refactored it in the final commit.

PR #2125, opened 2026-06-24
feat: Add swipe-to-delete-word functionality

skorokithakis:
I've added an option that, when swipe typing is off, will allow the user to swipe left (from anywhere) to delete the last word. This is something SwiftKey has and that I find very useful. Please let me know if this works!


skorokithakis:
I've signed but got no success or error message.

PR #2131, opened 2026-06-25
feat: Add spacebar language-swipe sensitivity setting

skorokithakis:
The spacebar language-switch swipe used a hardcoded 128dp step. I exposed it as an adjustable distance on the Spacebar settings screen with an 128dp default.

### DIFF
--- a/java/res/values/strings-uix.xml
+++ b/java/res/values/strings-uix.xml
@@ -440,6 +440,11 @@
     <string name="morekey_settings_spacebar_hold_shortcut">Hold</string>
     <string name="morekey_settings_spacebar_hold_shortcut_cursor">Move cursor</string>
     <string name="morekey_settings_spacebar_hold_shortcut_language">Switch language</string>
+    <string name="morekey_settings_spacebar_language_swipe_distance">Language swipe distance</string>
+    <string name="morekey_settings_spacebar_language_swipe_distance_subtitle">Shorter distance makes the language swipe more sensitive</string>
+
+    <!-- Units abbreviation for a distance in density-independent pixels [CHAR LIMIT=10] -->
+    <string name="abbreviation_unit_dp"><xliff:g id="DPI">%s</xliff:g>dp</string>
 
     <!-- Types of morekeys that can be configured, and examples of them -->
     <string name="morekey_settings_kind_numbers">Numbers</string>
--- a/java/src/org/futo/inputmethod/keyboard/PointerTracker.java
+++ b/java/src/org/futo/inputmethod/keyboard/PointerTracker.java
@@ -19,6 +19,7 @@ package org.futo.inputmethod.keyboard;
 import android.content.res.Resources;
 import android.content.res.TypedArray;
 import android.os.SystemClock;
+import android.util.DisplayMetrics;
 import android.util.Log;
 import android.view.MotionEvent;
 
@@ -97,10 +98,6 @@ public final class PointerTracker implements PointerTrackerQueue.Element,
     private static PointerTrackerParams sParams;
     private static final int sPointerStep = (int)(16.0 * Resources.getSystem().getDisplayMetrics().density);
     private static final int sPointerBigStep = (int)(32.0 * Resources.getSystem().getDisplayMetrics().density);
-    private static final int sPointerHugeStep = Integer.min(
-            (int)(128.0 * Resources.getSystem().getDisplayMetrics().density),
-            Resources.getSystem().getDisplayMetrics().widthPixels * 3 / 2
-    );
 
     private static GestureStrokeRecognitionParams sGestureStrokeRecognitionParams;
     private static GestureStrokeDrawingParams sGestureStrokeDrawingParams;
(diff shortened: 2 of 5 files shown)


FarisZR:
i think this has a weird edge case though, it's possible to change layouts multiple times in one swipe.
i think most users will only ever want it to change one layout per swipe, and still  want it to get triggered really easily.

Because now I either have a low enough sensitivity so that never happens (but that's a trade-off because i can't make it trigger as fast as i want) or i just have to not swipe for too long.

skorokithakis:
Ahh you're right, thanks, I'll test and make it only switch once.

FarisZR:
this patch makes it work as a latch, so that it only switches one layout at time

```patch
diff --git a/java/src/org/futo/inputmethod/keyboard/PointerTracker.java b/java/src/org/futo/inputmethod/keyboard/PointerTracker.java
--- a/java/src/org/futo/inputmethod/keyboard/PointerTracker.java
+++ b/java/src/org/futo/inputmethod/keyboard/PointerTracker.java
@@ -156,6 +156,7 @@ public final class PointerTracker implements PointerTrackerQueue.Element,
     private boolean mStartedOnFastLongPress;
     private boolean mCursorMoved = false;
     private boolean mSpacebarLongPressed = false;
+    private boolean mSpacebarLanguageSwipeTriggered = false;
 
     // true if keyboard layout has been changed.
     private boolean mKeyboardLayoutHasBeenChanged;
@@ -757,6 +758,7 @@ public final class PointerTracker implements PointerTrackerQueue.Element,
             mStartTime = System.currentTimeMillis();
             mStartedOnFastLongPress = key.isFastLongPress();
             mSpacebarLongPressed = false;
+            mSpacebarLanguageSwipeTriggered = false;
 
             mIsSlidingCursor = key.getCode() == Constants.CODE_DELETE || key.getCode() == Constants.CODE_SPACE;
             mIsFlickingKey = !mIsSlidingCursor && key.getHasFlick();
@@ -973,28 +975,36 @@ public final class PointerTracker implements PointerTrackerQueue.Element,
                         || (!mSpacebarLongPressed && settingsValues.mSpacebarSwipeMode != Settings.SPACEBAR_MODE_OFF);
 
             if(allowedBySettings) {
+                final boolean isLanguageSwipe = settingsValues.mSpacebarSwipeMode == Settings.SPACEBAR_MODE_LANGUAGE
+                        && !mSpacebarLongPressed;
                 int pointerStep = sPointerStep;
-                if (settingsValues.mSpacebarSwipeMode == Settings.SPACEBAR_MODE_LANGUAGE && !mSpacebarLongPressed) {
+                if (isLanguageSwipe) {
                     // The language-switch step distance is user-configurable in dp. The
                     // widthPixels*3/2 clamp is a safety bound for narrow screens.
                     final DisplayMetrics displayMetrics =
                             Resources.getSystem().getDisplayMetrics();
                     pointerStep = Integer.min(
                             (int)(settingsValues.mSpacebarLanguageSwipeStepDp * displayMetrics.density),
                             displayMetrics.widthPixels * 3 / 2
                     );
                 }
 
                 int steps = (x - mStartX) / pointerStep;
                 final int swipeIgnoreTime = settingsValues.mKeyLongpressTimeout / MULTIPLIER_FOR_LONG_PRESS_TIMEOUT_IN_SLIDING_INPUT;
                 if (steps != 0 && mStartTime + swipeIgnoreTime < System.currentTimeMillis()) {
-                    mCursorMoved = true;
-                    mStartX += steps * pointerStep;
-
-                    if (settingsValues.mSpacebarSwipeMode == Settings.SPACEBAR_MODE_LANGUAGE && !mSpacebarLongPressed) {
-                        sListener.onSwipeLanguage(steps);
+                    if (isLanguageSwipe) {
+                        if (!mSpacebarLanguageSwipeTriggered) {
+                            mCursorMoved = true;
+                            mSpacebarLanguageSwipeTriggered = true;
+                            sListener.onSwipeLanguage(steps > 0 ? 1 : -1);
+                        }
                     } else {
+                        mCursorMoved = true;
+                        mStartX += steps * pointerStep;
                         sListener.onMovePointer(steps);
                     }
                 }


```

skorokithakis:
I've implemented the gating on a single language. I'm not sure what the original intent was (to be able to switch multiple languages or not), but it seems more confusing to be able to switch more than one language, so I've added the constraint here, especially since the added sensitivity makes it very easy to overshoot.

abb128:
Do you think it would make more sense to adjust the default rather than adding an option for this? It feels like it would be a strangely specific setting

skorokithakis:
Depends on how the maintainers want it.

abb128:
@skorokithakis @FarisZR What values have you guys settled on for the setting?

skorokithakis:
72 DPI for me.

FarisZR:
I have it set to the minimum, so 24dp, because now we don't have that bug that switches multiple times in one swipe and i want it to react asap.

Still doesn't feel as seamless as something like Swiftkey, but it's much better now.

-------- Original Message --------

abb128:
Please try nightly 0.1.29.1-3-g9f05e86ac and tell me what you think. I added some visual and haptic feedback, made the switch more instantaneous, and reduced the threshold to 64dp

FarisZR:
64dp is still too much imo, at least from trying out this PR, i haven't tested your version yet

>

PR #2134, opened 2026-06-26
Collapse multilingual subtypes by keyboard layout

### LINKED ISSUE #1172: Option to Remove Language Switcher with Multilingual Typing Enabled
### Use Cases

If all languages have multilingual enabled you don't have to switch between multiple languages, would be nice if you could simple have one keyboard layout for all languages and no button to switch between layouts.

### Proposal

I can think of to ways:
1. If all languages have multilingual typing enabled and the keyboard layouts match (qwerty = qwerty) it should merge the keyboard layouts and hide the switch button.
2. currently every language has to have at least one keyboard layout, if multilingual is enabled you could remove that layout on your secondary language.

either way, there should be a button to hide the language name on the space bar

### References

_No response_

### LINKED ISSUE #1533: Feature request: Full multilingual support
### Use Cases

Copied from the discord (not something I wrote, I just also want the feature):

> Hello, I want to ask about something I've seen it talked but didn't see any workaround. Multilingual typing feature is great, however I do not need an English keyboard as I can write in English just fine from my native language's layout. It gets in the way when switching to another layout. FYI, I'm on Japanese branch

and:

> By default its the language I want, but clicking the language button swipes through 3 layouts since English layout is also there. I know I can directly select by holding space bar, but that is quite inconvenient as I change languages a lot.

### Proposal

_No response_

### References

_No response_

FarisZR:
combine languages that use the same layout and have multilingual typing enabled.

(PR written by 5.5 extended thinking)

## Technical details

- Treat switchable subtypes with the same layout as one entry
- Show combined multilingual labels on the spacebar and subtype picker
- Keep multilingual locale lists scoped to the active keyboard layout

fix #1172 (it removes the duplicate layout)
fix #1533

### DIFF
--- a/java/src/org/futo/inputmethod/keyboard/MainKeyboardView.java
+++ b/java/src/org/futo/inputmethod/keyboard/MainKeyboardView.java
@@ -52,6 +52,7 @@ import org.futo.inputmethod.keyboard.internal.NonDistinctMultitouchHelper;
 import org.futo.inputmethod.keyboard.internal.SlidingKeyInputDrawingPreview;
 import org.futo.inputmethod.keyboard.internal.TimerHandler;
 import org.futo.inputmethod.latin.AudioAndHapticFeedbackManager;
+import org.futo.inputmethod.latin.RichInputMethodManager;
 import org.futo.inputmethod.latin.uix.DynamicThemeProvider;
 import org.futo.inputmethod.latin.R;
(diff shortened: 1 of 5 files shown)


FarisZR:
i tested this and it works as expected, however i think i just caught a bug.
swipe typing crashes the keyboard.

FarisZR:
that should've fixed it, just tested it.

PR #2141, opened 2026-06-28
Fix default selection in more-keys popup

### LINKED ISSUE #2140: Bug: Missing or delayed default selection in long-press diacritic pop-up
### Reproduction steps

The keyboard's touch event handler fails to consistently assign the default active state to the primary diacritic character within the pop-up menu. The issue is intermittent.

1. Tap an input field.
2. Perform a long-press (długie naciśnięcie) on the 's' key.
3. Observe the pop-up menu (menu podręczne) initialization.
4. Release the screen touch.
5. Repeat steps multiple times, as the bug triggers randomly.

Attached video demonstrating the intermittent lack of selection state in the pop-up menu.

### Actual result

The auto-focus triggers with a delay or fails completely. Releasing the finger without an explicit swipe registers the base character 's'. This disrupts fast typing by forcing manual visual verification.

https://github.com/user-attachments/assets/7f8e250e-4309-47ba-b76a-98790c50742e

### Expected result

The UI should instantly highlight the primary alternative character ('ś') upon invoking the long-press (długie naciśnięcie). Releasing the finger should input 'ś' by default.

### Keyboard Version

0.1.29.1

Android 16

Nothing Phone 3

Nothing OS

### References

_No response_

minisiowo:
Fixes #2140 

## Problem

The long-press more-keys popup currently initializes its active key using coordinate-based hit testing based on the last touch position. If the translated touch position misses the popup key area due to touch jitter or event timing, the popup may open without an active (default) selection.

As a result, the default more key (for example, `ś` for Polish `s`) may not be selected immediately, and releasing the finger without moving can fail to commit the expected alternative character.

## Solution

This change preserves the existing coordinate-based selection behavior while adding a fallback mechanism:

* the more-keys keyboard now exposes its default key
* after the synthetic popup `onDownEvent()`, the panel selects the default key if no key is currently selected
* drag-based selection continues to work normally after the popup is opened

## Testing

Manually tested on a physical device:

* Polish layout
* long-press `s`
* verified that `ś` is selected immediately when the popup appears
* verified that releasing without dragging commits `ś`
* repeated the test multiple times; the intermittent missing-selection issue no longer occurred

### DIFF
--- a/java/src/org/futo/inputmethod/keyboard/MoreKeysKeyboard.java
+++ b/java/src/org/futo/inputmethod/keyboard/MoreKeysKeyboard.java
@@ -35,14 +35,20 @@ import javax.annotation.Nonnull;
 
 public final class MoreKeysKeyboard extends Keyboard {
     private final int mDefaultKeyCoordX;
+    private final Key mDefaultKey;
 
     MoreKeysKeyboard(final MoreKeysKeyboardParams params) {
         super(params);
         mDefaultKeyCoordX = params.getDefaultKeyCoordX() + params.mDefaultKeyWidth / 2;
+        mDefaultKey = params.mDefaultKey;
     }
 
     public int getDefaultCoordX() {
         return mDefaultKeyCoordX;
+    }
+
+    public Key getDefaultKey() {
+        return mDefaultKey;
     }
 
     @UsedForTesting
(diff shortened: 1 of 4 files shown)


minisiowo:
Hi, the CLA has been signed. Could you please update the PR status?

PR #2179, opened 2026-07-24
fix: add missing Serbian and Montenegrin letters

savely-krasovsky:


savely-krasovsky:
Done.

savely-krasovsky:
This should also cover Croatian, Bosnian and Slovene.

Issue #2184, opened 2026-07-26
Feature request: Settings home screen icons are visually inconsistent; Miscellaneous and Credits have no icon


personn-b:
### Use Cases

On the settings home screen, most rows have a coloured circular icon on the left — Languages & Models, Keyboard & Typing, Swipe Typing, Text Prediction, Voice Input, Actions, Theme, Payment, Help & Feedback.

Three rows don't: Miscellaneous, Credits, and Check for updates manually.

Because the rows without icons still keep the same left indent for their text, they leave an empty gap where every other row has a coloured circle. Scanning down the list, the eye follows the column of icons and then loses the thread at the bottom, so those three entries read as less important than the rest, or as though something failed to load. They're also harder to pick out at a glance, since there's no shape or colour to anchor them — only text.

Screenshot:

<img width="1080" height="2793" alt="Image" src="https://github.com/user-attachments/assets/be17b6fa-8c0a-4545-b03d-eb18e23cf687" />

### Proposal

Give the three rows icons, consistent with the existing ones.

In `java/src/org/futo/inputmethod/latin/uix/settings/pages/Home.kt`, all three are declared without an `icon` parameter:

- `Miscellaneous` — `NavigationItemStyle.MiscNoArrow`
- `Credits` — `NavigationItemStyle.MiscNoArrow`
- `Check for updates manually` — `NavigationItemStyle.Misc`

One thing worth deciding up front: `Misc` and `MiscNoArrow` set `circleColor = Color.Transparent` and dim the tint to 75% alpha, so simply adding `icon =` would give a faded icon with no circle behind it. Matching the rows above means also switching to `HomePrimary`/`HomeSecondary`/`HomeTertiary` — but that brings back the trailing arrow, which `MiscNoArrow` suppresses. If you'd rather those rows stay arrow-less, a new enum entry would be needed instead. Happy to go either way.

For the icons themselves, `more_horizontal.xml` already exists and fits Miscellaneous. Credits and Check for updates would need new assets — the closest existing candidates are already used elsewhere.

**Would you be open to a PR for this?** And where would you prefer new icons come from? The current set looks like Feather Icons, which is no longer maintained. Options I can see:

- [Lucide](https://lucide.dev) — the community fork of Feather, same 24×24 grid and 2 px stroke, ISC licensed, so new icons would match the existing assets exactly. Has `users` and `refresh-cw`.
[...]

PR #2185, opened 2026-07-26
Settings home: make icons visually uniform

personn-b:
Partial follow-up to #2184, which was closed as not planned. This covers only the "existing icons don't match each other" half — Miscellaneous, Credits, and Check for updates are untouched and still have no icon, so none of the `NavigationItemStyle` / trailing-arrow questions from that issue apply here.

The current set is Feather-derived, which is unmaintained, and several rows had bespoke drawables drawn off-grid against it. Lucide is the maintained Feather fork on the same 24×24 grid and 2px stroke, so path data drops in exactly.

## Icons

| Row | Drawable | Was | Now |
| --- | --- | --- | --- |
| Languages | `globe.xml` | custom 26×26 globe | `globe` |
| Keyboard & Typing | `keyboard.xml` | custom key-grid icon, off-grid (62×36 viewport) | `keyboard` |
| Swipe | `swipe_icon.xml` | custom gesture icon, off-grid (32×34 viewport) | `line-squiggle` |
| Predictive Text | `text_prediction.xml` | custom lightbulb-ish icon | `spell-check-2` |
| Voice Input | `mic_fill.xml` | custom filled mic | `mic` |
| Payment | `dollar_sign.xml` | plain dollar sign | `circle-dollar-sign` |
| Theme | `themes.xml` | Feather-era palette | `palette` |

`smile.xml`, `help_circle.xml`, and `code.xml` were already exact Lucide matches and are untouched.

Only drawable file contents changed — no `R.drawable.*` references in `Home.kt` were renamed, so there are no other call sites to update. That does leave `mic_fill.xml` holding an outline mic; filenames were kept to avoid churn, but happy to rename if you'd prefer.

## Also outside the home screen

Two more rows had the same off-grid Feather-era problem and got the same treatment:

| Row | Drawable | Was | Now |
| --- | --- | --- | --- |
| Predictive Text page | `icon_spellcheck.xml` | off-grid (24×25 viewport, strokeWidth 1.52381) spellcheck glyph | `spell-check` |
| Clipboard pin action | `push_pin.xml` | filled solid pin | `pin` (outline, matches the stroke-based style used elsewhere) |

I audited the rest of the app for the same pattern (keyboard layout picker, action bar glyphs, emoji category tabs, theme editor icons, etc.) and there's a longer list of candidates, but scoped this PR down to just these two plus the home screen — the rest are either filled/branded intentionally, have no clean Lucide equivalent, or are borderline enough to warrant their own discussion rather than bundling here.

## Styling

Separate commit — drop it if you'd rather not.

Row styles currently mix `HomePrimary` / `HomeSecondary` / `HomeTertiary`, so the icon circles rotate through three theme hues. This sets them all to `HomeSecondary`. I think it reads calmer, but it's a taste call and I'm not attached to it — the icon commit stands alone without it.

## Testing

- All modified vector XML validated with `xmllint`
- `./gradlew :assembleDebug` clean, no new resource warnings
- Installed and checked on a Pixel 10 emulator

## Screenshots

| Before | After |
| --- | --- |
| Before | After |

Happy to close if this isn't wanted.

*AI-assisted: description drafting and vector conversion. Built and verified manually.*


abb128:
Thank you for the PR, but I think it might be easier for me to change any icons myself if needed. I don't think all of them should be changed. I will have a look into Lucide since I wasn't aware of it before, but some of these icons were custom-made. Also I don't think we should make everything in home the same color because it removes a bit of intentional visual distinction between the items

Issue #1790, opened 2025-12-29
Feature request: Per-Language Theme as a Visual Indicator for Multilingual Use


drdanielbender:
### Use Cases

I use the FUTO keyboard in a multilingual setup. While using the keyboard, especially when using voice transcription, I frequently realize too late that the wrong language is active. Being able to set a theme per language as a visual indicator would be incredibly helpful to prevent this.

PR #2187, opened 2026-07-26
Add configurable spacebar colors per language

drdanielbender:
## Summary

Addresses #1790.

Issue #1790 requests per-language themes as a visual indicator for multilingual use. This implements a focused solution to the underlying problem: each configured language can have its own spacebar color.

Coloring only the spacebar provides a clear, persistent indication of the active language while preserving the user's selected keyboard theme and avoiding the complexity of maintaining a complete theme for every language.

## Changes

- Add a spacebar color setting for each configured language.
- Provide preset colors and custom six- or eight-digit hexadecimal colors.
- Apply the configured color when the active language changes.
- Refresh the spacebar immediately when its color setting changes.
- Select a contrasting foreground color for the language label.
- Keep new labels and color names in translatable string resources.

## Screenshots

| Per-language setting | Color picker | Active keyboard |
| --- | --- | --- |
| Per-language spacebar color setting | Spacebar color picker | German keyboard with orange spacebar |

## Validation

- `./gradlew --no-daemon compileUnstableDebugJavaWithJavac`
- `git diff --check upstream/master...HEAD`


abb128:
I like the idea, but these colors seem a bit too bright, I think it should act as a more subtle tint instead of a fully bright color and look good across light and dark themes. I'm not sure if we can make it work well with custom themes, for example the Christmas theme. I also think the hex color input should be removed, as well as the color names, for simplicity (the names will likely be difficult to translate to some languages)

drdanielbender:
## Update after review

Implemented the requested UI and color adjustments in `b1e92a16d`:

- Added a per-language transparency slider so the selected color can act as a subtle tint across light and dark themes. The default transparency is 50%.
- Removed the custom hex-color input.
- Removed the visible color names from the picker while retaining accessible descriptions.
- Replaced the pink preset with white.

As expected by @abb128 in [his comment](https://github.com/futo-org/android-keyboard/pull/2187#issuecomment-5093194003), the Christmas theme is not supported by this change and does not work with it. From my point of view, this is acceptable because not everyone will use the optional feature that changes the spacebar color based on the selected language. If someone likes to use this feature, they will likely be fine with using another theme.

| Per-language setting | Color picker | Active keyboard |
| --- | --- | --- |
| Per-language spacebar color setting | Spacebar color picker with transparency control | Active keyboard with tinted spacebar |

PR #2199, opened 2026-08-02
Add configurable swipe-to-delete words gesture

djurcola:
## Summary
- add an enabled-by-default Swipe settings option for a deliberate left swipe on ordinary keyboard keys
- select preceding words proportionally to swipe distance and delete the selection on release
- preserve swipe typing, space/delete gestures, flick keys, multi-touch, and normal taps

## Tests
- `./gradlew testUnstableDebugUnitTest compileUnstableDebugSources --console=plain`

## Notes
The gesture excludes modifier, flick, space, and delete keys so their existing behaviours remain unchanged.


djurcola:
Here are some screenshots of the implementation:

Swipe delete toggle

Swipe delete hold

PR #2200, opened 2026-08-02
Add optional GIPHY GIF action

djurcola:
## Summary
- add a GIF action that users can place through the Action Editor
- provide trending/search results through GIPHY only after the user supplies their own API key
- cache GIF downloads and insert them using the IME image-content URI path
- expose download and unsupported-editor failures in the action UI

## Privacy and scope
- no API key is included in this change
- the action is limited to the unstable flavor, where it declares the required network permission

## Tests
- `./gradlew testUnstableDebugUnitTest compileUnstableDebugSources --console=plain`

This is intentionally separate from the swipe-delete PR because it introduces an external online service.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -352,7 +352,7 @@ android {
 
         unstable {
             java.srcDirs = ['common/src', 'java/src', 'java/stable/java']
-            manifest.srcFile 'java/stable/AndroidManifest.xml'
+            manifest.srcFile 'java/unstable/AndroidManifest.xml'
             res.srcDirs = ['java/unstable/res', translationsWithoutEngValues('translations/devbuild')]
         }
 
--- a/java/res/drawable/gif.xml
+++ b/java/res/drawable/gif.xml
@@ -1,0 +1,34 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+    android:width="24dp"
+    android:height="24dp"
+    android:viewportWidth="24"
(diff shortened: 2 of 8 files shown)


djurcola:
Implementation screenshots:

Gifs API
Gifs In Action
Gifs Action2
Gifs Action

abb128:
Hi, thank you for the PR but the [decision was already made long ago to not include GIF search in FUTO Keyboard](https://docs.keyboard.futo.tech/troubleshooting/faq#when-will-gif-search-be-added). Giphy in particular makes mention of targeted advertising in their privacy policy, and requires the developer to make the user aware that their data will be collected for advertising:
> **Advertising.** Licensee acknowledges that Giphy may make advertising available via its API.  Such advertising will appear in connection with the content and/or services available via the SDK and Licensee agrees to follow all applicable advertising disclosure guidelines contained in the SDK.  Licensee acknowledges and agrees that Giphy will collect certain data from end users of the Application through the SDK as further set forth in its [Privacy Policy](https://support.giphy.com/hc/en-us/articles/360020028332-GIPHY-Privacy-Policy), and Licensee shall be solely responsible for all notices to its end users regarding the same.

This is just not what we need in FUTO Keyboard and would dilute the point of the product, which is to be an offline keyboard app that stands against giants who engage in this behavior. There's no way to make something like this optional while keeping the product coherent

PR #2201, opened 2026-08-03
Add configurable flick gesture sensitivity setting

rishubil:
## Summary

The flick gesture threshold is currently hardcoded to one third of the key width in `Key.flickDirection`. On flick layouts with small keys, or for users who prefer shorter gestures, that distance is farther than comfortable, and there is no way to adjust it.

This adds a "Flick sensitivity" slider so the threshold can be tuned, while keeping the current behavior as the default.

## Changes

- Add a "Flick sensitivity" slider to the "Long-Press Keys & Spacebar" settings menu, placed next to the existing "Long Press Duration" slider.
- Store the threshold as an integer percentage of the key width, adjustable from 1% to 100%, with `-1` selecting the built-in default of one third.
- Replace the hardcoded `width / 3` threshold in `Key` with a configurable `Key.flickThresholdRatio` companion property.
- Read the initial value and observe subsequent changes in `LatinIME`.

The slider follows the existing "Vibration" setting: the stored unit matches the slider range, values are quantized on write so the indicator always matches what is stored, and the `-1` sentinel renders as "Default".

## Notes

- Default behavior is unchanged — the threshold stays at one third of the key width unless the user moves the slider.
- Only affects layouts that use flick keys (the Japanese 12-key layouts, and custom flick layouts).

### DIFF
--- a/java/res/values/strings-uix.xml
+++ b/java/res/values/strings-uix.xml
@@ -425,6 +425,10 @@
     <string name="morekey_settings_show_hints_tags">additional characters, alt keys, symbols, popup keys</string>
     <string name="morekey_settings_duration">Long Press Duration</string>
     <string name="morekey_settings_duration_subtitle">How long a key needs to be pressed to be considered a long-press</string>
+    <string name="morekey_settings_flick_threshold">Flick sensitivity</string>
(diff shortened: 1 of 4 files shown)


rishubil:
I've signed the CLA

Issue #23, opened 2024-06-13
Feature Request: Capitalize word with cursor at the end of the word


Zetaphor:
Coming from Swiftkey, I'm finding a huge usability miss when trying to upper/lowercase a word

If I type a word and the cursor is at the end of that word, I would like if the shift button (capitalize button?) uppercased the word, and inversely if the word is already uppercased, lowercase it

Alexanderkocha:
I would love to see something similar to this

I believe this would be annyoying for words with unusual capitilasation, like PuTTY, for exaple. Maybe the smartbar can be used to suggest the change in case instead of applying with just the press of the shift key? 

The keyboard should temporarly remember the exact lower/uppercase characters of the original, incase user wants to revert to the original. This can also be shown in the smartbar

Progaros:
> Coming from Swiftkey, I'm finding a huge usability miss when trying to upper/lowercase a word
> 
> If I type a word and the cursor is at the end of that word, I would like if the shift button (capitalize button?) uppercased the word, and inversely if the word is already uppercased, lowercase it

you can already double tap the word to highlight it and press shift to Capitalize or UPPERCASE it

Alexanderkocha:
> > Coming from Swiftkey, I'm finding a huge usability miss when trying to upper/lowercase a word
> > If I type a word and the cursor is at the end of that word, I would like if the shift button (capitalize button?) uppercased the word, and inversely if the word is already uppercased, lowercase it
> 
> you can already double tap the word to highlight it and press shift to Capitalize or UPPERCASE it

It would be nice to see another way of capitalising a word only by interacting with the keyboard, in addition to being able to highlight it

WhiteBahamut:
It would really improve the usability if we can do it with keyboard-only. After switching from swiftkey, this is the feature I miss most!

dean-futo:
#474 and #235 are duplicates - merging those requests into this one.

nicr9:
One thing I've noticed since switching to FUTO keyboard is that I frequently get suggestions for a word and the same word with a capital letter. If we had this feature to quickly capitalize, there would be no need for duplicate predictions and it could free up space for another word prediction.

I realize that would probably increase the scope of this issue though...

Ziggi76:
I'm trying hard on the change from swiftkey to futo, this is by far the most missing feature.

Please consider implementation.

Thx

la-la-lars:
[...]

PR #2205, opened 2026-08-05
Recapitalize word at cursor with Shift

### LINKED ISSUE #1587: Feature request: Capitalize word at cursor
### Use Cases

I know this feature from Swiftkey.

When pressing Shift,  it capitalises the word where the cursor is at. 

### Proposal

_No response_

### References

_No response_

djurcola:
## Summary
- recapitalize the word at or adjacent to the cursor when Shift is pressed without a selection
- retain the existing selected-text recapitalization cycle
- preserve the input-start safety guard for selected-text recapitalization while validating the cursor-only path directly
- accept a recapitalized word before Space or punctuation is inserted, preventing both stale auto-correction and replacement of the recapitalization selection
- add coverage for word boundaries, Unicode, repeated Shift cycling followed by Space, and punctuation/whitespace no-op cases

Fixes #1587

## Validation
- `git diff --check`
- Android APK build and release completed successfully in the related custom integration build: https://github.com/djurcola/android-keyboard/actions/runs/31008152288
- Device regression test pending upstream review/build: type `tgis`, press Shift once or twice, then Space; the selected casing must remain and the separator must be inserted after it.

### DIFF
--- a/common/src/org/futo/inputmethod/latin/common/StringUtils.java
+++ b/common/src/org/futo/inputmethod/latin/common/StringUtils.java
@@ -405,6 +405,49 @@ public final class StringUtils {
         return true;
     }
 
+    /**
+     * Finds the word at or immediately before the cursor.
+     *
+     * The word is the maximal run of code points which are not in sortedWordSeparators that
+     * contains the cursor, starts right at the cursor, or ends right at the cursor.
+     *
+     * @param textBeforeCursor the text ending right at the cursor position.
+     * @param textAfterCursor the text starting right at the cursor position.
+     * @param sortedWordSeparators a sorted array of word separator code points.
+     * @return int[]{index where the word starts in textBeforeCursor,
+     *         index just past where the word ends in textAfterCursor}, or null if the cursor
+     *         is not adjacent to a word.
+     */
+    @Nullable
+    public static int[] getWordRangeAtCursor(@Nullable final CharSequence textBeforeCursor,
+            @Nullable final CharSequence textAfterCursor,
+            @Nonnull final int[] sortedWordSeparators) {
+        final int beforeLength = null == textBeforeCursor ? 0 : textBeforeCursor.length();
+        int start = beforeLength;
+        while (start > 0) {
+            final int codePoint = Character.codePointBefore(textBeforeCursor, start);
+            if (Arrays.binarySearch(sortedWordSeparators, codePoint) >= 0) {
+                break;
+            }
+            start -= Character.charCount(codePoint);
+        }
+        int end = 0;
+        if (null != textAfterCursor) {
+            final int afterLength = textAfterCursor.length();
+            while (end < afterLength) {
+                final int codePoint = Character.codePointAt(textAfterCursor, end);
+                if (Arrays.binarySearch(sortedWordSeparators, codePoint) >= 0) {
+                    break;
+                }
+                end += Character.charCount(codePoint);
+            }
+        }
+        if (start == beforeLength && end == 0) {
+            return null; // The cursor is not adjacent to a word on either side.
+        }
+        return new int[] { start, end };
+    }
+
     // TODO: like capitalizeFirst*, this does not work perfectly for Dutch because of the IJ digraph
     // which should be capitalized together in *some* cases.
     @Nonnull
--- a/java/src/org/futo/inputmethod/latin/inputlogic/InputLogic.java
+++ b/java/src/org/futo/inputmethod/latin/inputlogic/InputLogic.java
@@ -1971,12 +1971,11 @@ public final class InputLogic {
      * @param settingsValues The current settings values.
      */
     private void performRecapitalization(final SettingsValues settingsValues) {
-        if (!mRecapitalizeStatus.mIsEnabled()) return;
-
         int selectionStart = mConnection.getExpectedSelectionStart();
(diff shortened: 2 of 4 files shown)


la-la-lars:
I just tried the dev build and wanted to give feedback: while this is a godsend for typing in German, it makes typing something voluntarily in camelCase practically impossible because you cannot continue appending a capital letter to a word.
Long-pressing the Shift key or a gesture involving Shift would be better IMO.

djurcola:
> I just tried the dev build and wanted to give feedback: while this is a godsend for typing in German, it makes typing something voluntarily in camelCase practically impossible because you cannot continue appending a capital letter to a word. Long-pressing the Shift key or a gesture involving Shift would be better IMO.

Long press is a great idea! I will implement that.

la-la-lars:
> Long press is a great idea! I will implement that.

Actually, long- press might be not that great thinking about cycling through the capitalization variants (which is absolutely useful), as it incurs multiple delays. A gesture might be a better solution. With the old Swype/Dragon keyboard it was a short swipe gesture from a neighboring key to the shift key. Here, that could be swiping from the numeric switch button to the shift key.

Felix14-v2:
> typing something voluntarily in camelCase practically impossible

Alternative solution: detect compound words and apply camelCase to them as a capitalization option

la-la-lars:
> Alternative solution: detect compound words and apply camelCase to them as a capitalization option

I'm not the owner of this PR and can't speak for @djurcola, but IMHO that would not be an alternative but another feature altogether (another capitalization style), while this is about just a more convenient way of invoking the already known capitalization.

djurcola:
I am testing a few options:
1. Hold shift to cycle through options. Caps lock changed to double tap.
2. Swipe right - I use swipe left to delete the last word so this works good for me.
3. Swipe up. This one is a bit tricky on my height keyboard, I find it gets confused and inserts a number over the word instead. 

Currently enjoying 1 and 2 the most. 

Check my custom build on my profile if you want to try it.

https://github.com/djurcola/android-keyboard/actions/runs/31540741039

All the options are under swipe keyboard settings currently, will need to clean that up.

la-la-lars:
Just tried it. Holding shift works much better than expected thanks to the delay being short enough 👍
But: All the swipe left/right/up options conflict with general swipe typing (which I normally use and the biggest reason to use FUTO) though, even to the degree that the keyboard just crashes when both are enabled. Even if I disable all three, I get a crash on swipe typing 🙁
My suggestion was a local swipe gesture specifically from the special character switch key to the shift key. But now I'm not entirely sure if that would be better than holding shift.

djurcola:
I will remove those from the pull request, I have them since I don't use swiping but do like swipe left to delete!

I will adjust the functionality to hold shift as that is what I have used the most. Swipe from shift key was too precise for my fat fingers and I always missed the button.

EmilJunker:
I would like to propose an alternative implementation that avoids the issue with camelCase altogether. How about when the user types the word "hello" and then presses the shift key while the cursor is still adjacent to the word, Futo simply shows "Hello" in the suggestion bar. And when the user presses shift again, it shows "HELLO" in the suggestion bar. This allows the user to easily capitalize the word without interfering with the typed text or the ability to write camelCase.

la-la-lars:
My 2 cents: while @EmilJunker 's suggestion would work, it would in most cases double needed key strokes. Think long german combined words, which are mostly nouns, first suggested change is simple capitalizarion. Hold shift = one click and no need to look for the desired variant VS hold shift + select the desired variant = two clicks and felt broken flow. At least unless combined with option _suggestion bar mode = 2 alternatives_ which I strongly dislike!
Re camelCaae: neither a problem with hold shift interaction (just type camel case and keep it), nor would I think it worth an extra variant - you wouldn't want to start there and continue with PascalCase, kebab-case, .....

Therefore, I'd rather keep @djurcola 's idea.

la-la-lars:
> I will remove those from the pull request, I have them since I don't use swiping but do like swipe left to delete!
> 
> I will adjust the functionality to hold shift as that is what I have used the most. Swipe from shift key was too precise for my fat fingers and I always missed the button.

@djurcola once you can fix the swiping conflict I'd be happy to test a new dev build 😊

Then, is there anything one can do to help getting the PR accepted?

djurcola:
Hey Guys, it is pretty hard to get a feature merged here so I am going to leave the current build as is, which has all options available (and able to be toggled on off) for now. I don't think they will approve the hold shift as that is currently mapped to caps.

abb128:
I'll reopen this to look at when I get around to it but I wasn't very happy about some of the proposed behavior, namely that recapitalizing skips autocorrect. Over the past month I've been occupied with the voice input app so I've not had the chance to review this properly unfortunately. Sorry about the silence.

djurcola:
No worries @abb128 - but the current PR is not suitable for merging, it was just a test of a few different features / options and is a bit of a mess. If there is one method that is preferred (everyone seems to like the hold shift) I can clean it up and update the PR.

la-la-lars:
> I wasn't very happy about some of the proposed behavior, namely that recapitalizing skips autocorrect. 

I'd like to note that doing recapitalization _without_ applying autocorrect may well be desired - see example for compound German nouns in [here](https://github.com/futo-org/android-keyboard/issues/23#issuecomment-5480529635).

PR #2225, opened 2026-08-14
Treat Russian ё as a language long-press key

shamsartem:
## Summary

Treat Russian `ё` as a language long-press key.

`ё` is rarely used, so users should be able to put numbers before it. It has a similar role to Polish `ę`, so both should follow the Language Keys order.

## Testing

- `assembleUnstableDebugAndroidTest`


shamsartem:
@Zvonimir-FUTO is there some way to speed up review of this. It's only one line and a safe change. Will make people's life a bit easier so they can just choose to type the number 5 by default like on most of other keyboards I ever used

abb128:
Thanks!

PR #2226, opened 2026-08-15
Add suggested words to the user dictionary

djurcola:
## Summary

Adds **Add to dictionary** to the long-press menu for normal suggested words.

- Uses the existing Android `UserDictionary` integration for the active locale.
- Avoids duplicate no-shortcut entries and refreshes active suggestions after insertion.
- Keeps **Blacklist** available for normal suggestions.
- Preserves the existing emoji menu: **Blacklist** and **Disable emoji suggestions** only.

## Screenshot

Normal suggested-word long press on device:

Long-press menu with Add to dictionary

## Validation

- `git diff --check` passed.
- Resource XML parses successfully.
- The behavior was exercised in the downstream implementation on a physical Android device.
- Local Gradle unit testing is blocked before compilation in this environment because no Android SDK is configured; upstream CI is expected to provide the Android build validation.

The screenshot is hosted on a dedicated media-only branch of the contributor fork and is not part of this PR's code diff.

### DIFF
--- a/java/res/values/strings-uix.xml
+++ b/java/res/values/strings-uix.xml
@@ -182,9 +182,14 @@
     <string name="autoupdater_perform_update">Update</string>
 
     <!-- These strings can appear in the keyboard itself -->
+    <string name="keyboard_suggest_manage_word_body">Manage the word \"<xliff:g name="wordToManage" example="crane">%s</xliff:g>\"</string>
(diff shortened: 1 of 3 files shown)


abb128:
I'm not a fan of having this for a few reasons
1. It makes the menu way more cluttered and confusing. People may misclick on one while trying to do the other
2. There's already an automatic word learning feature. This pushes people to manually managing word dictionary which is not the recommended way of using the keyboard

PR #2258, opened 2026-08-26
fix: isExactMatch: strip diacritics for accent-aware exact matching

YannBirba:
This PR was created by human (me), but all the code LLM generated.
I hope it's okay for you.

## Problem

When the GGML language model suggests an accented word (e.g. `Garçon`) for a
user who typed an unaccented ASCII input (`garcon`), the exact-match scoring
rule at `org_futo_inputmethod_latin_xlm_LanguageModel.cpp:1273-1286` fires
incorrectly:

```cpp
// Exact match rule
bool hasExactMatch = false;
for(const auto &result : results) {
    if(isExactMatch(result.second, partialWordString)) {
        hasExactMatch = true;
    }
}
if(hasExactMatch){
    for(auto &result : results) {
        if(!isExactMatch(result.second, partialWordString)) {
            result.first -= 1.0f;  // <-- penalizes accented candidates
        }
    }
}
```

`isExactMatch` previously only lowercased and stripped `'`, `-`, and spaces.
UTF-8 accented characters (`é`, `ç`, `ê`, …) were compared byte-for-byte,
so `"Garçon"` never matched `"garcon"`. The -1.0 penalty pushed accented
candidates below the unaccented form, defeating the LM's accent predictions.

## Root cause

`isExactMatch` did not normalize Unicode diacritics to their ASCII base forms
before comparison. A user typing `garcon` on an ASCII keyboard would always
see `Garcon` instead of `Garçon`, even when the LM correctly predicted the
accented form.

## Fix

Add `append_diacritic_normalized()`, a lightweight UTF-8 normalizer that maps
accented characters to their base ASCII equivalents:

- **Latin-1 Supplement** (U+00C0-U+00FF, `0xC3` prefix): all common
  European accents (French, Spanish, Italian, Portuguese, German, Nordic)
  — including uppercase variants (À-Ÿ) and ligatures (Æ→"ae").
- **Latin Extended-A**: Œ/œ (`0xC5 0x92`/`0x93`) → "oe".

Unmapped sequences (Ð, Þ, ß, …) keep their original bytes, preserving
stock matching behavior.

## Testing

Added `native/jni/tests/test_isexactmatch.cpp` — a standalone test
(compilable with `g++ -std=c++17`) covering:

| Category | Examples |
|----------|----------|
| French lowercase accents | garçon↔garcon, café↔cafe, être↔etre |
| Mixed case | Garçon↔GARCON, CAFÉ↔CAFE |
| Ligatures | cœur↔coeur, Œuvre↔Oeuvre |
| Separators stripped | l'homme↔lhomme, Hello-World↔helloworld |
| Negative cases | maison≠maisonn, chat≠chien |
| Unmapped sequences | ðtest≠dtest (preserves stock behavior) |

```
33/33 tests passed
```

Also tested on-device (Android emulator, AZERTY layout) with the custom
GGML model:

| Typed | Expected | Result |
|-------|----------|--------|
| garcon ␣ | Garçon | ✓ |
| Xafe ␣ | Café | ✓ |
| bonjor ␣ | Bonjour | ✓ |
| etre ␣ | Être | ✓ |
| francais ␣ | Français | ✓ |

## Scope & limitations

- Covers Latin-1 Supplement + Œ/œ. Other scripts (Cyrillic, Arabic, CJK)
  are untouched and preserve stock byte-for-byte comparison.
- `ñ` normalizes to `n` for exact-match purposes (intentional: beam search
  should not penalize `ñ` when user typed `n`).
- Trade-off: words differing only by accent (a/à, ou/où) will both count
  as exact matches. The LM probability still decides ranking; auto-correction
  thresholds guard unintended replacements.

### DIFF
--- a/native/jni/org_futo_inputmethod_latin_xlm_LanguageModel.cpp
+++ b/native/jni/org_futo_inputmethod_latin_xlm_LanguageModel.cpp
@@ -154,13 +154,75 @@ bool hasLowercase(const char* str, bool strict) {
     return false;
 }
 
+// Appends the ASCII normalization of the UTF-8 sequence starting at str[i]
+// to out, and returns the number of bytes consumed (0 if the sequence is
+// not an accented character we normalize). Unmapped sequences keep their
+// original bytes, preserving stock matching behavior for them.
+static int append_diacritic_normalized(const unsigned char *str, size_t len, size_t i, std::string &out) {
+    if(i + 1 >= len) return 0;
+    unsigned char c1 = str[i];
+    unsigned char c2 = str[i + 1];
+
+    // Latin-1 Supplement (U+00C0-U+00FF and uppercase counterparts):
+    // 2-byte UTF-8 sequences starting with 0xC3.
+    if(c1 == 0xC3) {
+        switch(c2) {
+            case 0x80: case 0x81: case 0x82: case 0x83: case 0x84: case 0x85: // À Á Â Ã Ä Å
+            case 0xA0: case 0xA1: case 0xA2: case 0xA3: case 0xA4: case 0xA5: // à á â ã ä å
+                out += 'a'; return 2;
+            case 0x86: case 0xA6:                                             // Æ æ
+                out += "ae"; return 2;
+            case 0x87: case 0xA7:                                             // Ç ç
+                out += 'c'; return 2;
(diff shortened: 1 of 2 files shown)


YannBirba:
Just signed the CLA

Issue #2288, opened 2026-09-07
Bug: The delete button for thai language should only delete a glyph not the whole column


peperronii:
### Reproduction steps

1. Choose Thai keyboard layout
2. Type some thing like this:  มี or ผุ
3. Press the delete button

### Actual result

มี and ผุ will get deleted entire ly

### Expected result

มี should turn into ม
ผุ should turn into ผ

### Keyboard Version

0.1.30 ver code 11751

### Which android version are you using?

Android 17

### Which device are you using?

Xperia 10 mk VI

### Which operating system are you using?

Android

peperronii:
By the way, Loas and Burmese will also have this issue and all of these arise from the fact that getCharacterInstance returns the length of the grapheme cluster.

Edit: on further research, this will affect ALL Complex Script.

PR #2290, opened 2026-09-08
pass the current locale to getWordInstance and getCharacterInstance

peperronii:
Fix issue [#2288](https://github.com/futo-org/android-keyboard/issues/2288) and potentially other languages that has a combined glyph rendering.


peperronii:
After manual testing, the behaviour for the fix is unstable. I'll go fix this first before requesting a review.

peperronii:
I've checked. This is a problem with the ICU's getCharacterInstance implementation. But, we should still pass the locale to the call any way, so I'll leave it like this

peperronii:
Sorry for the confusion. I've read the ICU's spec and misunderstood its "character" defenition. The implementation is correct and if we fix this we must actually use the same logic as the Korean language. ( some how on the first debug run, it does fix the issue :P )

PR #2296, opened 2026-09-10
Handle Complex Scripts properly on handleBackspaceEvent

### LINKED ISSUE #2288: Bug: The delete button for thai language should only delete a glyph not the whole column

peperronii:
Fix #2288 , ~~The Handle Emoji comment is also misleading because it is already handled by the above Character.isSupplementaryCodePoint~~ -> This is not true. I just knew we have multiple-codepoint emojis. This needs rework. -> rework DONE!

## Validation

- build successful
- manual tested on my Xperia 10 VI
- Tried to run the tests located in InputLogicTests.java but an exception was raised (help is very much appreciated, I'd like to add the test case for this fix if the test is working)


peperronii:
This deletes only the unicode codepoint for complex scripts and for multiple-codepoint emoji it deletes the whole cluster. I used the UProperty.EMOJI and UProperty.EMOJI_COMPONENTS which will bump the minimum requirement to [Android API level 28](https://developer.android.com/reference/android/icu/lang/UProperty#EMOJI)

